PR #1, opened 2016-06-01
Add Dockerfile and make docker target

riywo:
This PR is for adding a feature to build a binary within a docker container. It is useful if the customer don't want to setup Go environment for building.

Note: If you use a remote Docker Machine, this won't work since the local volume is not mounted to the container directly.


samuelkarp:
Thanks!

PR #3, opened 2016-06-28
Adding local file cache for credential helper

sentientmonkey:
This provides a JSON file cache that's serialized/deserialized
for each access. The local file cache tokens are kept for the half-life
of the token lifetime (i.e. a 12 hour token will get expired after 6
hours). Users may also opt-out of local cache with an environment
variable set of `AWS_ECR_DISABLE_CACHE`.


ecr-login/cache/file.go
```diff
@@ -0,0 +1,140 @@
...
+		log.Infof("Could not clear cache: %s")
+	}
+}
+
+// Saves credential cache to disk. This writes to a temporary file first, then moves the file to the config location.
+// This elminates from reading partially written credential files, and reduces (but does not eliminate) concurrent
+// file access. There is not guarantee here for handling multiple writes at once since there is no out of process locking.
+func (f *fileCredentialCache) save(registryCache *RegistryCache) error {
+	defer log.Flush()
+	file, err := ioutil.TempFile("", "ecr-credential-cache")
```
samuelkarp: Can you change this to use a temp file in the home directory?  `os.Rename` will fail if the temp file and destination file are not on the same filesystem, so creating the temp file in the same directory would help that problem.

```go
dir, _ := homedir.Dir()
file, err := ioutil.TempFile(dir, "ecr-credential-cache")
```

samuelkarp:
Looks good other than the temp file bit.

ecr-login/cache/file.go
sentientmonkey: I did not know that. 😞  Seems reasonable to keep all within `~/.ecr`. I'll update, will inject homedir path into the file cache.

sentientmonkey:
Added temp file & rebased against master.

samuelkarp:
@sentientmonkey Looks like some unit test failures.

sentientmonkey:
Looks like it was a deep comparison issue with times. 😭 Comparing each of the values seemed to do the trick. Not exactly sure why this was failing in travis vs. locally.

samuelkarp:
:+1: 

I think my one remaining concern here is garbage collection of expired cache entries when the access key changes (like how it will rotate on an EC2 instance profile), but we can follow up with that separately.

sentientmonkey:
Will create issue for cleaning up expired entries - seems reasonable to purge expired entries on load or save.

sentientmonkey:
Added: https://github.com/awslabs/amazon-ecr-credential-helper/issues/6

Issue #7, opened 2016-08-02
Expired cached token is considered valid -> authorization fails


f-barth:
I'm seeing some weird behaviour in the credentials helper:
I have a token from yesterday in my cache:
```
# date -u +"%Y-%m-%dT%H:%M:%SZ"
2016-08-02T08:23:40Z

# cat .ecr/cache.json
{
  "Registries": {
    "<snip>": {
      "AuthorizationToken": "<snip>",
      "RequestedAt": "2016-08-01T17:05:07.865313389Z",
      "ExpiresAt": "2016-08-02T05:05:08Z",
      "ProxyEndpoint": "https://<snip>.dkr.ecr.eu-west-1.amazonaws.com"
    }
  },
  "Version": "1.0"
}

```

When I do a `docker pull` I get the following result:
```
#docker pull <snip>.dkr.ecr.eu-west-1.amazonaws.com/test:latest
Error response from daemon: error parsing HTTP 403 response body: invalid character 'Y' looking for beginning of value: "Your Authorization Token has expired. Please run 'aws ecr get-login' to fetch a new one."
```
 
The credentials helper's log shows the following:
```
2016-08-02T08:24:55Z [DEBUG] Retrieving credentials for <snip> in eu-west-1 (<snip>.dkr.ecr.eu-west-1.amazonaws.com)
2016-08-02T08:24:55Z [DEBUG] GetCredentials for <snip>
2016-08-02T08:24:55Z [DEBUG] Checking file cache for <snip>
2016-08-02T08:24:55Z [DEBUG] Using cached token for <snip>
```

It seems to me, as if the credentials helper does somehow not deem the expired token invalid. Do you have any insights on that?

PR #8, opened 2016-08-02
Fixes cache token expiry

f-barth:
After seeing the behaviour in #7, I dug into the cache expiration logic.

With `authEntry.ExpiresAt.After(testTime.Add(-1 * (window / time.Duration(2))))` we did prolong the the validity of the token rather than shortening it by window/2.


samuelkarp:
@f-barth Thank you!  I've taken your commits and attempted to make the logic more readable in #10.

f-barth:
@samuelkarp, 👍 on increased readability, thx!

samuelkarp:
@f-barth Can you confirm this contribution is under the terms of the Apache 2.0 License?  Thanks!

f-barth:
sorry for the delay :) I can confirm that.

PR #22, opened 2016-12-20
gogenerate: Exclude vendored go:generate lines

samuelkarp:
This excludes vendored go:generate lines from being executed by `make gogenerate`.


sentientmonkey:
:shipit:

PR #24, opened 2017-02-19
Updates README

iMerica:
I found the current information architecture of the README to be a little counter intuitive.

Since the README is in English and English reads from the top down, it makes sense to place the steps in order. Currently the installation steps come first, but they don't make sense unless one has already read the building steps, which come after.

I went ahead and merged the two sections and renamed it to "Building and Setup" because the user has build it themselves, not install it.

If/When the binary is published somwehere easily pulled down with a package manager, then it would be good to add a Installation section.


nrdlngr:
This reorganization looks good to me, too. This README could use an editing pass, but let's do that after this is merged; I don't want to hold it up for that.

samuelkarp:
@iMerica Thanks!

PR #26, opened 2017-02-24
Implement helper list

### LINKED ISSUE #23: docker build --pull issue with docker-engine 1.13
Can easily test with a simple dockerfile
```
FROM <id>.dkr.ecr.us-east-1.amazonaws.com/ubuntu14.04:latest
CMD ["echo", "test"]
```
Docker 1.13, build it with --pull option will fail.
```
docker build --pull -t test .
Sending build context to Docker daemon 2.048 kB
Step 1/2 : FROM <id>.dkr.ecr.us-east-1.amazonaws.com/ubuntu14.04:latest
Pulling repository <id>.dkr.ecr.us-east-1.amazonaws.com/ubuntu14.04
unauthorized: authentication required
```
Errors in /var/log/upstart/docker.log
```
time="2017-02-08T17:49:47.089538193Z" level=error msg="Attempting next endpoint for pull after error: Get https://<id>.dkr.ecr.us-east-1.amazonaws.com/v2/ubuntu14.04/manifests/latest: no basic auth credentials"
time="2017-02-08T17:49:47.104752428Z" level=error msg="Not continuing with pull after error: unauthorized: authentication required"
```
.docker/config.json:
```
{
    "credsStore": "ecr-login",
    "auths": {
        "<id>.dkr.ecr.us-east-1.amazonaws.com": {
        }
    }
}
```
Tested several workarounds:
- docker 1.12 has no problem
- docker 1.13, removing --pull option, docker build is fine
- docker 1.13, removing "credsStore": "ecr-login" from config.json, use aws ecr get-login, docker build --pull is fine.

sentientmonkey:
Beginning on implementation for the List API to fix #23. This vends the latest docker-credential-helper and fixes the build by providing a notImplemented interface for List.

From looking at the List API implementation, this appears to return all service urls and usernames. We should be able to provide similar functionality by walking the cache file and returning all of the repositories and usernames within the auth tokens. Repositories not already cached will not be returned. Anyone using the null cache will never get any values returned as well. We should also be able to implement Erase, but Add still doesn't really make sense for our use cases.

The major area we'll have to refactor here is that the ClientFactory today is created regionally, and the cache really global. I think the right approach here is to expose the cache up one level higher for access by our helper. Thoughts? @samuelkarp @yinshiua

### DIFF
--- a/ecr-login/api/client.go
+++ b/ecr-login/api/client.go
@@ -30,6 +30,7 @@ const proxyEndpointScheme = "https://"
 
 type Client interface {
 	GetCredentials(registry, image string) (*Auth, error)
+	ListCredentials() ([]*Auth, error)
 }
 type defaultClient struct {
 	ecrClient       ecriface.ECRAPI
@@ -98,6 +99,22 @@ func (self *defaultClient) GetCredentials(registry, image string) (*Auth, error)
 	return nil, fmt.Errorf("No AuthorizationToken found for %s", registry)
 }
 
+func (self *defaultClient) ListCredentials() ([]*Auth, error) {
+	log.Debug("Listing credentials from cache")
+
+	auths := []*Auth{}
+	for _, authEntry := range self.credentialCache.List() {
+		auth, err := extractToken(authEntry.AuthorizationToken, authEntry.ProxyEndpoint)
+		if err != nil {
+			log.Debugf("Could not extract token: %v:", err)
+		} else {
+			auths = append(auths, auth)
+		}
+	}
+
+	return auths, nil
+}
+
 func containsProxyEndpoint(image string, proxyEndpoint string) bool {
 	if image == "" {
 		return true
@@ -108,9 +125,13 @@ func containsProxyEndpoint(image string, proxyEndpoint string) bool {
 func extractToken(token string, proxyEndpoint string) (*Auth, error) {
 	decodedToken, err := base64.StdEncoding.DecodeString(token)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("Invalid token: %v:", err)
 	}
+
 	parts := strings.SplitN(string(decodedToken), ":", 2)
+	if len(parts) < 2 {
+		return nil, fmt.Errorf("Invalid token: expected two parts, got %n", len(parts))
+	}
 
 	return &Auth{
 		Username:      parts[0],
--- a/ecr-login/ecr.go
+++ b/ecr-login/ecr.go
@@ -15,7 +15,9 @@ package ecr
 
 import (
 	"errors"
+	"fmt"
 	"regexp"
+	"strings"
 
 	"github.com/awslabs/amazon-ecr-credential-helper/ecr-login/api"
 	log "github.com/cihub/seelog"
@@ -68,6 +70,24 @@ func (self ECRHelper) Get(serverURL string) (string, string, error) {
 }
 
 func (self ECRHelper) List() (map[string]string, error) {
-	//TODO implement for docker 1.13 --pull
-	return nil, notImplemented
+	log.Debug("Listing credentials")
+	client := self.ClientFactory.NewClientWithDefaults()
+
+	auths, err := client.ListCredentials()
+	if err != nil {
+		log.Errorf("Error listing credentials: %v", err)
+		return nil, fmt.Errorf("Could not list credentials: %v:", err)
+	}
+
+	result := map[string]string{}
+
+	for _, auth := range auths {
+		serverURL := removeEndpointProtocol(auth.ProxyEndpoint)
+		result[serverURL] = auth.Username
+	}
+	return result, nil
+}
+
+func removeEndpointProtocol(endpointWithProtocol string) string {
+	return strings.TrimPrefix(endpointWithProtocol, "https://")
 }
--- a/ecr-login/ecr.go
+++ b/ecr-login/ecr.go
@@ -31,6 +31,9 @@ type ECRHelper struct {
 	ClientFactory api.ClientFactory
 }
 
+// ensure ECRHelper adheres to the credentials.Helper interface
+var _ credentials.Helper = (*ECRHelper)(nil)
+
 func (ECRHelper) Add(creds *credentials.Credentials) error {
 	// This does not seem to get called
 	return notImplemented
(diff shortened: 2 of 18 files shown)


samuelkarp:
> We should also be able to implement Erase, but Add still doesn't really make sense for our use cases.

Add would be fine if it just adds it to the cache though.  I think that's okay.

> I think the right approach here is to expose the cache up one level higher for access by our helper.

Sounds good.

ecr-login/ecr.go
```diff
@@ -31,6 +31,9 @@ type ECRHelper struct {
 	ClientFactory api.ClientFactory
 }
 
+// ensure ECRHelper adheres to the credentials.Helper interface
+var _ credentials.Helper = (*ECRHelper)(nil)
```
samuelkarp: Neat!

sentientmonkey: Yeah, actually from go docs - [godoc/#guarantee_satisfies_interface](https://golang.org/doc/faq#guarantee_satisfies_interface). I usually avoid except in cases like this where we want to keep clear that we fully implement interface. It does blow up in cmd/main.go but this makes it much more explicit.

sentientmonkey:
Rebased against @yinshiua's latest changes and took a first pass at the List implementation. Right now, it walks the FileCredentialsCache for any previously cached registries and returns them (filtered by region and credentials). This isn't quite right because it should really return registry urls for all regions - I think I'm going to have to update the cache prefix assumptions to fix. Good news is that it seems to fix the build case in docker 1.13 with the assumption that you've pulled or pushed that repository before so it's in our file cache.

Example:
```
$ cat Dockerfile.test
FROM <id>.dkr.ecr.us-west-2.amazonaws.com/busybox:latest
CMD ["echo", "test"]

$ docker build --pull -f Dockerfile.test .
Sending build context to Docker daemon 12.28 MB
Step 1/2 : FROM <id>.dkr.ecr.us-west-2.amazonaws.com/busybox:latest
latest: Pulling from busybox
Digest: sha256:1ded4559e2aab2ab3464aae5170ef64afc15bea324a0861b543e2c56a3f29711
Status: Image is up to date for <id>.dkr.ecr.us-west-2.amazonaws.com/busybox:latest
 ---> e02e811dd08f
Step 2/2 : CMD echo test
 ---> Using cache
 ---> 9818ecf1cb6e
Successfully built 9818ecf1cb6e
```
I should be able to fix the region-specific filtering, as well as implement Add & Erase, but wanted to send out what I have for feedback so others could test and see how it works so far, as well as checkpoint for review.

samuelkarp:
> Currently this relies on current AWS region and credentials to filter (not sure if this is a correct assumption to make).

I don't think we want to make that assumption, so we should remove the region/account filtering prior to merging this.

ecr-login/api/client.go
```diff
@@ -98,6 +99,22 @@ func (self *defaultClient) GetCredentials(registry, image string) (*Auth, error)
 	return nil, fmt.Errorf("No AuthorizationToken found for %s", registry)
 }
 
+func (self *defaultClient) ListCredentials() ([]*Auth, error) {
+	log.Debug("Listing Credentials from cache")
+
+	auths := []*Auth{}
+	for _, authEntry := range self.credentialCache.List() {
```
samuelkarp: Since you're just walking the cache, are you going to return expired credentials?  Should we refresh credentials for those that are expired so that valid entries can be returned?

samuelkarp:
Filtering by account that obtained the credentials might make sense, but the current cache scheme relies on the access key rather than the account.  When this is run under an assumed role, the access key can change but the set of registries for which it is authorized may remain the same.

ecr-login/api/client.go
sentientmonkey: We are returning expired credentials here - list just returns the serverUrls and usernames (not credentials).

```
$ /usr/local/bin/docker-credential-ecr-login list
{"<id>.dkr.ecr.us-west-2.amazonaws.com":"AWS"}
```

In the build case, Docker does a list first, then seems to call get for each one. That seems pretty reasonable and much simpler than refreshing ahead of time (since the get will refresh any that are expired). That way if the build process ever gets smarter, it will only refresh what it needs (vs. everything from list).

Totally agree about the filtering for regions - I want to fix that before merging - will require a bit of refactoring in the cache so it's not too much of a hack. The assumed role case isn't one I had thought of, but I wanted to scope down based on access (for the multiple profile case). The cacheKey is "region-hashedAccessKey-registry-id", so we could either move to 1) filtering by the hashedAccessKey only, 2) move from hashedAccessKey to hashedAccountId, and filter by that (if available from SDK), or 3) return all registries without any filtering. Thoughts?

samuelkarp: I *think* the best thing to do here is to just return everything then.  The Docker CLI will take care of requesting credentials by calling get at that point, which will handle returning the appropriate credentials.

If you want to scope it down further, you could look at by account.  But that will require making a call to [GetCallerIdentity](http://docs.aws.amazon.com/STS/latest/APIReference/API_GetCallerIdentity.html) that you might not have permission to make, at which point you'll probably want to fall-back to returning everything anyway.

sentientmonkey: I think that's the right call (returning everything). Worse case, we end up making get calls for registries that a profile hasn't accessed already, but the GetAuthorizationToken will succeed in all of these cases. Since registryId isn't a resource we conditionally evaluate, that means that you would have a token fetched for every single registry you've used in past. That's much better than the alternative of not being able to fetch a registry after switching back and forth.

I'll go ahead and go down that route and wrap this up earlier in that case (Add & Erase can wait for another commit), so we can get this fix out earlier for customers :smile:

sentientmonkey:
Ok, removed filtering and added a new default constructor for the API client that defaults the region (so it's a bit more encapsulated).

With the latest change:
```
./bin/local/docker-credential-ecr-login list
{"<id>.dkr.ecr.us-east-1.amazonaws.com":"AWS","<id>.dkr.ecr.us-west-2.amazonaws.com":"AWS"}
```

I'm gonna hold off on Add & Erase for now (so we can get this merged and get customers unblocked).

mikelorant:
Have confirmed this works.

rpnguyen:
@sentientmonkey lgtm :shipit:

PR #28, opened 2017-03-06
Get default registry's credential when List if cache doesn't exist

gengmao:
This is based upon https://github.com/awslabs/amazon-ecr-credential-helper/pull/26. The purpose is to make `docker build --pull` can still work even cache doesn't exist, hence fully address https://github.com/awslabs/amazon-ecr-credential-helper/issues/23. Basically, in the ListCredentials() function I invoked getAuthorizationToken for default registry if there's no credential cache. It only requires the user has set AWS_REGION environment variable.

Also refactored a bit the code:
- Client.GetCredentials() just needs one serverURL parameter. It can parse registry and region. 
- Client.GetCredentials() can auto remove the `https://` prefix from serverURL.
- Client.ListCredentials() and ECR.List() can return proxyEndpointUrl which has `https://` prefix.
- Extracted common logics of GetCredentials() and ListCredentials() into a getAuthorizationToken() function.
- Moved some codes from ecr.go to api/client.go.
- Adjusted tests and mocks accordingly. 

@sentientmonkey kindly review.

### DIFF
--- a/ecr-login/api/client.go
+++ b/ecr-login/api/client.go
@@ -32,23 +32,31 @@ const programName = "docker-credential-ecr-login"
 
 var ecrPattern = regexp.MustCompile(`(^[a-zA-Z0-9][a-zA-Z0-9-_]*)\.dkr\.ecr\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.amazonaws\.com(\.cn)?`)
 
-func ExtractRegistryAndRegion(serverURL string) (string, string, error) {
+type Registry struct {
+	ID	string
+	Region	string
+}
+
+func ExtractRegistry(serverURL string) (*Registry, error) {
 	if (strings.HasPrefix(serverURL, proxyEndpointScheme)) {
 		serverURL = strings.TrimPrefix(serverURL, proxyEndpointScheme)
 	}
 	matches := ecrPattern.FindStringSubmatch(serverURL)
 	if len(matches) == 0 {
-		return "", "", fmt.Errorf(programName + " can only be used with Amazon EC2 Container Registry.")
+		return nil, fmt.Errorf(programName + " can only be used with Amazon EC2 Container Registry.")
 	} else if len(matches) < 3 {
-		return "", "", fmt.Errorf(serverURL + "is not a valid repository URI for Amazon EC2 Container Registry.")
+		return nil, fmt.Errorf(serverURL + "is not a valid repository URI for Amazon EC2 Container Registry.")
 	}
-	registry := matches[1]
-	region := matches[2]
-	return registry, region, nil
+	registry := &Registry{
+		ID:	matches[1],
+	        Region:	matches[2],
+	}
+	return registry, nil
 }
 
 type Client interface {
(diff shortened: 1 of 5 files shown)


sentientmonkey:
Hi Mao, thanks for PR! Seems like a good approach - will review code. Are you ok contributing this code under Apache 2.0 license?

ecr-login/api/client.go
```diff
@@ -27,9 +28,27 @@ import (
...
+		return "", "", fmt.Errorf(serverURL + "is not a valid repository URI for Amazon EC2 Container Registry.")
+	}
+	registry := matches[1]
+	region := matches[2]
+	return registry, region, nil
+}
 
 type Client interface {
-	GetCredentials(registry, image string) (*Auth, error)
+	GetCredentials(serverURL string) (*Auth, error)
```
sentientmonkey: We may need to retain the existing interface here if this is being used outside the project (and just call down to the older API). @yinshiua are you using GetCredentials to pull in the CLI?

ecr-login/api/client.go
```diff
@@ -27,9 +28,27 @@ import (
...
+	}
+	matches := ecrPattern.FindStringSubmatch(serverURL)
+	if len(matches) == 0 {
+		return "", "", fmt.Errorf(programName + " can only be used with Amazon EC2 Container Registry.")
+	} else if len(matches) < 3 {
+		return "", "", fmt.Errorf(serverURL + "is not a valid repository URI for Amazon EC2 Container Registry.")
+	}
+	registry := matches[1]
+	region := matches[2]
+	return registry, region, nil
```
sentientmonkey: Minor: Could we add a new struct to return here that holds the registry and region?

ecr-login/api/client.go
yinshiua: Since CLI code is still in PR, I can adapt the change of function names, but this will make it non-reusable for the CLI. The CLI only knows about the registryID when calling GetCredentials, and gets the region from aws config.

gengmao:
Hi Scott, 
> Are you ok contributing this code under Apache 2.0 license?
Yes. It is fine to me. Is there an agreement to sign electrically? Please let me know.

ecr-login/api/client.go
gengmao: Done

ecr-login/api/client.go
gengmao: I extracted out another GetCredentialsByRegistryID(registryID string) method, which is re-used by GetCredentials(serverURL string). @yinshiua can you use the GetCredentialsByRegistryID(registryID string) method too?

yinshiua: Yes. GetCredentialsByRegistryID will work for the CLI! Thanks!

gengmao:
@sentientmonkey Bump :)

sentientmonkey:
No need to sign anything - you agreement above stating you agree that this change is under Apache 2.0 is good for us :+1:

sentientmonkey:
LGTM :shipit: Thanks again!

gengmao:
@sentientmonkey Fantastic. Thanks for reviewing!

PR #30, opened 2017-03-13
Add proxy support for retrieve ECR credential

lilida:
Hi,

In this pull request, I made some changes to support proxy scenario for ecr-credential-helper. Could you help on reviewing and let me know if it can be merged to the master tree? I think  it will be useful for a bunch of scenarios.

- What I did
I add a support to have a reverse proxy before ECR. A reverse proxy is a pull through cache for http requests.  For ex, the reverse proxy can be ecrproxy.myaws.com, the user run docker pull ecrproxy.myaws.com/repo:sha256xx . The reverse proxy will pull from ECR, forward to users and cache the data. This is useful in several scenarios for performance and resiliency.  However, to make this credential helper work, it needs to tell it the ECR url for the credential to retrieve. Right now, it only supports ECR url format

- How I did it
A ecr_proxy.json file that contains a map between proxy url and ecr url like below.

  {"proxies":{"ecrproxy.myasws.com":"12345678910.dkr.ecr.us-east-1.amazonaws.com"}}
Then when processing GET request. Credential helper will check if it can find the key in the map. If so, it uses the REAL url and the workflow goes through

- How to verify it
Tests I have done:
1. Function test: Set up a reverse proxy called ecrproxy.myasws.com (with nginx in this case but it can be any reverse proxy) and forward requests to ecr account. Run docker command lines.

2. Regression tests: Run docker command to directly pull from ECR and everything works.

Docker version: 1.12.5 OS: Ubuntu 14.04

### DIFF
--- a/ecr-login/config/proxy_config.go
+++ b/ecr-login/config/proxy_config.go
@@ -1,0 +1,55 @@
+package config
+
+import (
+	"os"
+	log "github.com/cihub/seelog"
+	homedir "github.com/mitchellh/go-homedir"
+	"path/filepath"
+	"encoding/json"
+)
+/**
+  Proxy config is used for storing a url to ECR registry mapping.
+  This is useful when setting up a reverse proxy for ECR to cache
+  container data.
+  {"proxies":{"ecrproxy.myasws.com":"12345678910.dkr.ecr.us-east-1.amazonaws.com"}}
+ */
+type ProxyConfig struct {
+	Proxies map[string]string `json:"proxies"`
+}
+
+func GetProxyConfig() (*ProxyConfig, error) {
+	config := ProxyConfig{}
+	dir, _ := homedir.Expand(GetCacheDir())
+
+	//Save it the same as the cache dir now
+	proxyFile := filepath.Join(dir, "ecr_proxy.json")
+	if _, err := os.Stat(proxyFile); err != nil {
+		log.Info("No Proxy config found")
+		return nil, err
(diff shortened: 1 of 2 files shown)


gengmao:
@lilida Are you trying to achieve `docker pull/push` to the proxy url rather than ECR url, and reference the image using proxy url in Dockerfile? 
There is a problem that is the List() will still return ECR url. Even if you can change it too, I think this feature is inappropriate to put into the ecr credential helper. 
Maybe it is better to build a new proxy-helper project base on the origin https://github.com/docker/docker-credential-helpers. It can import this ecr credential helper code, or just invoke the docker-credential-ecr-login and manipulate its stdin/stdout.

lilida:
@gengmao The purpose is only for pull. There is no intention for things like push. The only goal is to support  pull through cache work. I am not sure List() return ECR url is really concern of this case.

Since every registry might has its own way to support auth. Building a new helper looks to me are mostly repeated work of existing ECR helper. That is the reason that I want to merge it to the main one. I think pull through cache is useful in many cases and looks like we are not the only person trying to do this.

sentientmonkey:
@lilida While I'm happy that you were able to make this change for your use case, I'm not sure it's general enough to merge in and support within for all users of the credential helper. Setting up a proxy for ECR manually is not something that we have documentation or general support for today. I do encourage you to continue to use your fork for your use cases as needed and thank you for reaching out.

PR #43, opened 2017-06-28
Explain credHelpers doesn't work with docker-compose

danielcompton:
This section can be removed once docker-compose supports `credHelpers` again.


samuelkarp:
Thanks!

PR #44, opened 2017-07-05
Replace mockgen with testify mocks

lox:
The author of mockgen has basically recommended people avoid it: https://github.com/golang/mock/pull/35#discussion_r64802424

The generate approach is painful to get working and adds all sorts of orchestration dependencies and script requirements. This starts to slim things down by moving mocks into the codebase and using [testify/mock](http://godoc.org/github.com/stretchr/testify/mock) for the mock assertions.


samuelkarp:
Hi @lox, thanks for opening this.  Were you running into any particular problems with gomock and mockgen?  I looked at https://github.com/golang/mock/pull/35#discussion_r64802424, but neither of the people in that conversation appear to be authors of the gomock or mockgen, so I don't know that I would say that the author has recommended it to be avoided.

lox:
Hrm, I can't for the life of me find the thread I was leading that lead to that comment @samuelkarp. Reading through their issues list is a fairly cautionary tale though. It just strikes me as a lot of complexity for little return. In this PR we could even remove testify and have something that is much simpler to understand, debug and build.

I guess the flip side of it is, what value do you see gomock bringing for the extra complexity required around gogenerate?

samuelkarp:
> Reading through their issues list is a fairly cautionary tale though. It just strikes me as a lot of complexity for little return. 

Yep, that's fair.  I initially used gomock here because we use it in the ECS agent, and we use it in the ECS agent because (at the time) it looked like a good choice with respect to what was being used in the ecosystem.

> I guess the flip side of it is, what value do you see gomock bringing for the extra complexity required around gogenerate?

Inertia...it already works :smile:.  Ideally I'd like to be consistent in the projects I help maintain (this one, the ECS agent, ecs-init, the ECR containerd resolver, etc), and switching all of them would be a larger amount of work.

I'm not entirely opposed to switching, just trying to understand the motivation here.  If we do switch, I'd like to have something akin to mockgen so that the mocks don't need to be hand-maintained.

lox:
The thing is that hand-maintaining them means they are much smaller surface area and much easier to understand. We are talking about small interfaces that by virtue of the fact that they are public APIs just shouldn't change much.

These are the only golang projects I've encountered so far that use mocks. I'd argue that the additional complexity discourages casual contribution. Trying to make changes to the code required me to understand a lot of complexity just for stubbing out some methods in tests.

lox:
I'd encourage that you optimize for contribution friendly code rather than strong consistency with the other projects. The use of gomock doesn't seem like a pattern that is worth keeping as a standard to me.

lox:
I'll have a go at removing testify's mocking lib in another PR. If you are still keen on gomock, feel free to close both. I'll untangle the other PR's.

PR #45, opened 2017-07-05
Move godeps to govendor

lox:
This adds onto #44, as they were difficult to untangle. 

This moves to a top-level vendors dir managed by govendor. It will also drop support for 1.5 and adds support for 1.7 and 1.8.


samuelkarp:
Hi @lox, thanks for opening this.  Were you running into problems with godep?  Is there a reason you changed the vendor directory?  There are no go files in the top-level directory.

If we were to switch vendor tools, I'd be more inclined to use [dep](https://github.com/golang/dep) or [vndr](https://github.com/LK4D4/vndr).

lox:
Yup, think I agree it's not worth changing.

PR #47, opened 2017-07-05
Simpler mocks

lox:
This is another take on #44. It drops the gomock dependency for plain old go code. The interfaces under mocking are simple enough to just use stubs. This gives type safety (gomock doesn't always catch mismatched method signatures) and is radically simpler.


lox:
Shall I close this one out @samuelkarp?

PR #55, opened 2017-08-07
Added function to delete expired credentials from cache

### LINKED ISSUE #6: Remove expired tokens in file cache
On load or save, we should remove any expired tokens. Since we only use a token up to it's half-life, removing after expiration seems pretty reasonable. Since we additionally use a hash of the access key in the key, we may end up creating quite a number of entries over time.

richardpen:
This PR adds the functionality to clean up expired credentials at saving. Fix #6

### DIFF
--- a/ecr-login/cache/file.go
+++ b/ecr-login/cache/file.go
@@ -19,11 +19,15 @@ import (
 	"io/ioutil"
 	"os"
 	"path/filepath"
+	"time"
 
 	log "github.com/cihub/seelog"
 )
 
-const registryCacheVersion = "1.0"
+const (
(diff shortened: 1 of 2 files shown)


samuelkarp:
I think at this point we might want to skip this until https://github.com/moby/moby/issues/31517 is addressed.  The current behavior makes it so that `docker build` will work, as long as you've pulled from a registry at least once before.  If we start removing things from the cache, that behavior will break.

samuelkarp:
Closing for now.  If https://github.com/moby/moby/issues/31517 ever gets resolved, we can revisit.  For customers with a cache file that increases, they can delete the file.

PR #59, opened 2017-08-25
dockerfile: Use Go 1.9

samuelkarp:


samuelkarp:
Apparently Travis doesn't know about 1.9 yet.

PR #60, opened 2017-08-25
makefile: targets for different variants

samuelkarp:
@rpnguyen I can't seem to request a review from you here, but can you take a look too?

@nmeyerhans If you have any suggestions on cleaning up the Makefile I'm all ears. :smile:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -17,7 +17,12 @@ all: build
 
 SOURCEDIR=./ecr-login
 SOURCES := $(shell find $(SOURCEDIR) -name '*.go')
-LOCAL_BINARY=bin/local/docker-credential-ecr-login
+BINARY_NAME=docker-credential-ecr-login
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -38,6 +43,25 @@ $(LOCAL_BINARY): $(SOURCES)
 test:
 	. ./scripts/shared_env && go test -v -timeout 30s -short -cover $(shell go list ./ecr-login/... | grep -v /vendor/)
 
+.PHONY: all-variants
+all-variants: linux-amd64 darwin-amd64 windows-amd64
```
jahkeup: This is a great addition for this tool's build process :+1: 

With the general support for cross compilation in the official golang distribution as of 1.5, does it make sense to leave the `build` target instead of preferring to use `all-variants` in its place?

samuelkarp: I'd leave `build` as-is because I don't think everyone needs to build all variants all the time.

PR #72, opened 2017-10-03
Create standalone docker file & allow it to be used for retrieving credentials manually

surajbarkale-dolby:
This patch changes the Dockerfile so a proper docker image can be created. The binary within generated docker image is statically linked so it can be used in other Linux based Dockerfiles.

I also added ability to emulate behavior of `aws ecr get-login` so this docker image can be used to pull from ECR when only Docker is available (e.g. CoreOS)

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,10 +11,17 @@
 # ANY KIND, either express or implied. See the License for the specific
 # language governing permissions and limitations under the License.
 
-FROM golang:1.9
+FROM golang:1.9-alpine3.6 AS builder
 
-WORKDIR /go/src/github.com/awslabs/amazon-ecr-credential-helper
+COPY ecr-login /go/src/github.com/awslabs/amazon-ecr-credential-helper/ecr-login
 
-COPY . .
+RUN env CGO_ENABLED=0 go install -installsuffix "static" \
+    github.com/awslabs/amazon-ecr-credential-helper/ecr-login/cli/docker-credential-ecr-login
 
-CMD make
+FROM alpine:3.6
+
+RUN apk add --no-cache ca-certificates
+
+COPY --from=builder /go/bin/docker-credential-ecr-login /usr/local/bin/docker-credential-ecr-login
+
+ENTRYPOINT [ "/usr/local/bin/docker-credential-ecr-login" ]
--- a/Dockerfile
+++ b/Dockerfile
@@ -25,3 +25,4 @@ RUN apk add --no-cache ca-certificates
 COPY --from=builder /go/bin/docker-credential-ecr-login /usr/local/bin/docker-credential-ecr-login
 
 ENTRYPOINT [ "/usr/local/bin/docker-credential-ecr-login" ]
+CMD [ "eval" ]
--- a/ecr-login/cli/docker-credential-ecr-login/main.go
+++ b/ecr-login/cli/docker-credential-ecr-login/main.go
@@ -41,7 +41,7 @@ func evalCommand(helper credentials.Helper) {
 		var user, token string
 		user, token, err = helper.Get(server)
 		if err == nil {
-			fmt.Printf("docker login -e none -u %s -p %s %s\n", user, token, server)
+			fmt.Printf("docker login -u %s -p %s %s\n", user, token, server)
 		}
 	}
 	if err != nil {
--- a/ecr-login/cli/docker-credential-ecr-login/main.go
+++ b/ecr-login/cli/docker-credential-ecr-login/main.go
@@ -14,6 +14,9 @@
 package main
 
 import (
+	"fmt"
+	"os"
+
 	ecr "github.com/awslabs/amazon-ecr-credential-helper/ecr-login"
 	"github.com/awslabs/amazon-ecr-credential-helper/ecr-login/api"
 	"github.com/awslabs/amazon-ecr-credential-helper/ecr-login/config"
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -11,10 +11,18 @@
...
 
-WORKDIR /go/src/github.com/awslabs/amazon-ecr-credential-helper
+COPY ecr-login /go/src/github.com/awslabs/amazon-ecr-credential-helper/ecr-login
 
-COPY . .
+RUN env CGO_ENABLED=0 go install -installsuffix "static" \
+    github.com/awslabs/amazon-ecr-credential-helper/ecr-login/cli/docker-credential-ecr-login
 
-CMD make
+FROM alpine:3.6
```
samuelkarp: Is there any particular reason you need a full Alpine container instead of a scratch container?  The executable is statically-linked, so the only thing that should be necessary at that point is a CA certificate bundle rather than the rest of the Linux userland.

ecr-login/cli/docker-credential-ecr-login/main.go
```diff
@@ -24,5 +27,39 @@ import (
...
+	}
+}
+
+func evalCommand(helper credentials.Helper) {
+	server, err := parseArgs(helper)
+	if err == nil {
+		var user, token string
+		user, token, err = helper.Get(server)
+		if err == nil {
+			fmt.Printf("docker login -u %s -p %s %s\n", user, token, server)
```
samuelkarp: I'm a little leery of doing this.  Docker has already changed the arguments that it accepts for `docker login`; it used to be that `--email` was required, but now providing `--email` fails.  You can see the challenge that we had with the AWS CLI and adapting it here: https://github.com/aws/aws-cli/issues/1926.

Dockerfile
surajbarkale-dolby: Internally we used this container to run scripts within our CI so we found it useful to have basic UNIX utilities. Alpine overhead is around 4 MB which seemed reasonable to me. Also I am not sure how to just add the CA certificate bundle. If you have some pointers, I shall be happy to change this.

ecr-login/cli/docker-credential-ecr-login/main.go
surajbarkale-dolby: I know; the latest version of docker is recommending to pass the arguments over stdin using `--password-stdin` flag. It is very annoying that the CLI is not kept backwards compatible.

My primary use case was running `eval $(docker run --rm -e AWS_REGION=us-east-1 dolbylabs/amazon-ecr-credential-helper)` in our normal EC2 instances on startup to be able to login to ECR. Since we pull our images only at startup, this was sufficient.

The only other alternative is to vendor the `moby/client` and call [`RegistryLogin()`](https://godoc.org/github.com/moby/moby/client#Client.RegistryLogin) function directly. I can make this change if you no not mind the extra vendor dependencies. I can still use the resulting image by bind mounting docker socket.

samuelkarp:
Apologies for the delay.  Can you confirm that your contribution is under the terms of the Apache 2.0 license?

Dockerfile
samuelkarp: I think my preference would be for a scratch container so that we don't need to worry about updating the base layer.  You can see an example that builds a scratch container with a CA certificate bundle [here](https://github.com/jahkeup/updater53/blob/8a7f8bcd978d06241b96e0876fcc12c9c1d93d64/Dockerfile).

ecr-login/cli/docker-credential-ecr-login/main.go
samuelkarp: I don't think that using `moby/client`/`RegistryLogin()` with the Docker socket will actually cause the `~/.docker/config.json` file to be written since the Docker socket controls the Docker daemon and not the client.

I think your use-case makes sense, but I'm still concerned as `docker login` does change (and the change you mentioned I didn't even know about until you said it here!).  Maybe instead of just calling it `eval`, we can encode the expected Docker CLI version in the name of the command somehow?  Something like `eval-1.11` so that it produces a `docker login` command without `--email`?

surajbarkale-dolby:
I hereby confirm that all my contributions to this repository are under the terms of the Apache 2.0 license.

Dockerfile
surajbarkale-dolby: OK I shall update the Dockerfile to use scratch base.

ecr-login/cli/docker-credential-ecr-login/main.go
surajbarkale-dolby: I can change `eval` to `eval-1.11` so it is more explicit. I can also add `eval-17.06` with following code: 
  `fmt.Printf("echo %s | docker login --password-stdin -u %s %s", token, user, server)`

It will help to silence warnings from docker.

surajbarkale-dolby: The email flag for docker login has been optional since [docker v1.4](https://docs.docker.com/v1.4/reference/commandline/cli/#login). So I do not think we need to worry about the email flag. I can add `--password-stdin` flag to eval command to handle the new way and add `-e` flag without options to output email.

samuelkarp: I don't know offhand, but I'm pretty sure email was still required as of Docker 1.9.  I think I would prefer separate commands that correspond to the first version where the Docker CLI expected the invocation.

ecr-login/cli/docker-credential-ecr-login/main.go
```diff
@@ -24,5 +27,66 @@ import (
...
+		var user, token string
+		user, token, err = helper.Get(server)
+		if err == nil {
+			var emailOpt string
+			if email {
+				emailOpt = " -e none"
+			}
+			var echoPassword, passwordOpt string
+			if passwordStdin {
+				echoPassword = fmt.Sprintf("echo %s|", token)
```
samuelkarp: It might be better to do this as `<<< %s` so that `echo`'s command line entry in `/proc` doesn't have the token.

surajbarkale-dolby: Fish does not support herestrings. So I am hesitant in making this change.

PR #93, opened 2018-01-16
change `make docker` to copy from finished container instead of mounting

akdor1154:
Currently `make docker` works by mounting the output dir on the host into the build container.
This means it can't be used when the build environment is itself a docker container.

This PR changes the docker build behaviour to instead run the build container without mounting, and then copying the binary out of the finished container before removing it.

Benefits:
 - works when building inside container environments

Drawbacks:
 - slightly more complicated
 - because the build container is removed each time, `make docker` builds from scratch every time (if this is an issue it could be worked around with the drawback of more complexity in the Makefile)
 - requires Docker 1.8

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -24,13 +24,18 @@ LINUX_AMD64_BINARY=bin/linux-amd64/$(BINARY_NAME)
 DARWIN_AMD64_BINARY=bin/darwin-amd64/$(BINARY_NAME)
 WINDOWS_AMD64_BINARY=bin/windows-amd64/$(BINARY_NAME).exe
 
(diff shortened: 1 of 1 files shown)


simonbuchan:
Also required for Windows (at least using WSL) where the only way to use docker is `export DOCKER_HOST=tcp://localhost:2375`, e.g. "remote" connection to docker for windows.

I did pretty much exactly this except for `docker rm` at the end of the `docker:` target, which should match the behavior of the previous `docker run --rm`, the difference being that if `run`(?) or `cp` fails, the container is left over.

Shouldn't `docker rm -v` be used? Not sure if (unmounted) volumes are created and thus leaked.

Issue #84, opened 2017-12-02
failed to make it work with assume role


loivis:
I've read through related issues and it seems like to work well with assume role. However I just didn't manage to get it work. I would like to post my configuration for someone to help me out.

### Goal:
I have one iam user in `master` account, with permission to assume role in another `sub` account. The role in `sub` account has managed `AdministratorAccess` policy to all resources in that account. With `assume-role`, I would like the iam user in `master` account to pull/push docker images to ecr repository in `sub` account.

### Configuration:
+ ~/.aws/credentials
```
[default]
aws_access_key_id = xxx
aws_secret_access_key = xxx
```
+ ~/.aws/config
```
[default]
output = json
region = eu-west-1
cli_history = enabled

[profile sub]
role_arn = arn:aws:iam::221234567890:role/master-access
source_profile = default
region = eu-west-1
cli_history = enabled
```

+ ~/.docker/config.json
```
{
  "credsStore": "ecr-login"
}
```

### steps
+ with `default` profile
Hepler works with ecr in current account as I've always been using.
```
bash-3.2$ echo $AWS_PROFILE $AWS_DEFAULT_PROFILE $AWS_SDK_LOAD_CONFIG

bash-3.2$ docker pull 111234567890.dkr.ecr.eu-west-1.amazonaws.com/testing:latest
latest: Pulling from testing
Digest: sha256:a09e6318d74f23fe68bad84259fedf5e6ce02c38537336419ee5e9abe10aa2af
Status: Image is up to date for 111234567890.dkr.ecr.eu-west-1.amazonaws.com/testing:latest
```

+ switch to `sub` profile as assumed role from `master` to `sub`
```
bash-3.2$ export AWS_PROFILE=sub
bash-3.2$ export AWS_SDK_LOAD_CONFIG=true
bash-3.2$ echo $AWS_PROFILE $AWS_DEFAULT_PROFILE $AWS_SDK_LOAD_CONFIG
sub true
````
assume role works as I could list images in `sub` account
```
bash-3.2$ aws ecr list-images --repository-name oracle-java
{
    "imageIds": [
        {
            "imageDigest": "sha256:a9e3c92cdc3d8f8da92bf55ca1455098e3c6447e5ce6c64b15ad59cab072a98e",
            "imageTag": "8"
        }
    ]
}
```
But error occurs while trying to run `docker pull` to the same repository.
```
[...]

PR #98, opened 2018-02-07
vendor: update aws-sdk-go

samuelkarp:
Fixes https://github.com/awslabs/amazon-ecr-credential-helper/issues/84

### DIFF
--- a/ecr-login/Gopkg.lock
+++ b/ecr-login/Gopkg.lock
@@ -32,8 +32,8 @@
     "service/ecr/ecriface",
     "service/sts"
   ]
-  revision = "1b176c5c6b57adb03bb982c21930e708ebca5a77"
-  version = "v1.12.70"
+  revision = "586c9ba6027a527800564282bb843d7e6e7985c9"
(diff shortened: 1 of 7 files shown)


samuelkarp:
For the record, here's how I bisected to find the error and validated that the fix works:

```bash
#!/bin/bash

set -x

make
if [[ $? -ne 0 ]]; then
	exit 125
fi

# default profile can pull from its own repository
docker pull 123456789012.dkr.ecr.us-west-2.amazonaws.com/test:latest
if [[ $? -ne 0 ]]; then
	exit 125
fi

# default profile cannot pull from other repository
docker pull 987654321012.dkr.ecr.us-west-2.amazonaws.com/test:latest
if [[ $? -ne 1 ]]; then
	exit 125
fi

# assume-role profile can pull from other repository
AWS_SDK_LOAD_CONFIG=1 AWS_PROFILE=assume-role docker pull 987654321012.dkr.ecr.us-west-2.amazonaws.com/test:latest
exit $?

```

PR #100, opened 2018-02-10
Support "docker build"

abicky:
"docker build" fails if the base image is an image in ECR.
For example, if we try to build an image using Dockerfile
with a content like below:

```
From XXXXXXXXXX.dkr.ecr.ap-northeast-1.amazonaws.com/test
```

we will receive an error message like "repository
XXXXXXXXXX.dkr.ecr.ap-northeast-1.amazonaws.com/test
not found: does not exist or no pull access".

I think this change will resolve https://github.com/awslabs/amazon-ecr-credential-helper/issues/76.

### DIFF
--- a/ecr-login/ecr.go
+++ b/ecr-login/ecr.go
@@ -16,6 +16,7 @@ package ecr
 import (
 	"errors"
 	"fmt"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


abicky:
I tried to write steps to reproduce the issue but I couldn't reproduce it..., so I close this PR.

Issue #80, opened 2017-10-11
Tagged releases and binary artifacts


samuelkarp:
Hey all,

There have been a [few](https://github.com/awslabs/amazon-ecr-credential-helper/issues/15) [different](https://github.com/awslabs/amazon-ecr-credential-helper/issues/33) [issues](https://github.com/awslabs/amazon-ecr-credential-helper/issues/74) asking questions about publishing tagged releases and binaries.  I wanted to provide a list of what things need to occur before those can happen just to give a bit more context.  These are all things that we do for the other binaries released by ECS including the ECS agent and ECS CLI.

For tagged releases, we need to do the following:

* [x] Pick a versioning scheme and assign a first version.  Right now there is no version.
* [x] Expose a way to interactively query the version of the installed binary.  This is probably a `-v` or `--version` flag.  This can be contributed if anyone is looking to do so.
* [x] Override the default user-agent string used by the AWS SDK to include the name of the program and the version number.  We want this for tracking the usage of the credential helper vs. the AWS CLI or interacting with the API and to track different versions of the credential helper.  This can be contributed if anyone is looking to do so.
* [x] Determine release criteria for when a release should be tagged.

For publishing binaries, we need to do the following:

* [x] Pick which set of platforms that we'll build binaries for.  This includes both operating systems (Linux, Windows, macOS) and archtectures (arm32v6 for Raspberry Pi 1/Zero, for example).
  * [x] Ensure that behavior is sane on each platform.  Windows may need changes for canonical log storage locations, for example.  This can be contributed if anyone is looking to do so.
* [x] Pick a build system.  CodeBuild is pretty reasonable and there's a [pull request](https://github.com/awslabs/amazon-ecr-credential-helper/pull/61) for it already.
* [x] Determine where binaries should be vended.
* [x] Set up an auditing system for automatically verifying the integrity of published binaries.
* [x] Build out scripts and a process for creating releases with binaries.

Some of this work is easy enough for us to accept contributions, while other work must be accomplished by the ECS/R team.

Thanks,
Sam

lox:
I'm happy to implement the `--version` and user-agent stuff.  Seems like semver is a no-brainer and given the infrequency of change perhaps an ad-hoc release process would be a good way to start? 

Something like https://github.com/goreleaser/goreleaser would have releases done and dusted in under 60 minutes. Perhaps a build system and a verification system could be an incremental step? For a project that is infrequently updated, is it worth investing all this effort?

[...]

PR #105, opened 2018-04-30
Add version to cli and user-agent

srizzling:
Hi there!

I saw this #80, and thought I would provide an implementation of adding version to cli and useragent. I saw mentions of goreleaser which is definitely a good option to go for. So with that in mind, my PR allows easy integration with goreleaser, since you configure the LDFLAGS similar to the implementation made here.

I made a whole bunch of assumptions, so please excuse any mistakes in that regard, but I thought I would just implement something first and then we can easily discuss if it was the right approach.

Confusingly, `docker-credential-ecr-login` has a version command which setup in: https://github.com/docker/docker-credential-helpers/blob/f00de1b72fc8274c009c019a34d6bfade31c715c/credentials/version.go

So instead of `docker-credential-ecr-login version` the approach would be to go: `docker-credential-ecr-login -version` or `docker-credential-ecr-login -v`.

I think I got the userdata stuff working, but I'm kinda fresh to golang, and I couldn't read how to run the tests or add tests to see if the user agent was set. If you could help me out that would be cool 

Thanks,
Sriram


samuelkarp:
Hi @srizzling, thanks for sending this!  Can you confirm that your contribution is licensed under the Apache 2.0 license?

VERSION.txt
```diff
@@ -0,0 +1 @@
+0.0.1
```
samuelkarp: Would you mind renaming this file to `VERSION` to match what we do in our [other](https://github.com/aws/amazon-ecs-agent/blob/v1.17.3/VERSION) [projects](https://github.com/aws/amazon-ecs-cli/blob/v1.5.0/VERSION)?

Makefile
```diff
@@ -17,7 +17,8 @@ all: build
 
 SOURCEDIR=./ecr-login
 SOURCES := $(shell find $(SOURCEDIR) -name '*.go')
-BINARY_NAME=docker-credential-ecr-login
+VERSION := $(shell cat VERSION.txt)
+BINARY_NAME=docker-credential-ecr-login-$(VERSION)
```
samuelkarp: Can you revert this change to the `BINARY_NAME`?  The resulting binary name is still `docker-credential-ecr-login` in `build_binary.sh` (see line 37 of that file) and changing it here breaks `make`'s ability to track unnecessary rebuilds.  (With this change, `make linux-amd64` rebuilds every time you invoke it, but before this change `make` will tell you "make: Nothing to be done for 'linux-amd64'." when invoked multiple times.)


Also: Docker depends on the exact binary name and I'd like to keep it easy for people to just copy the binary into their `PATH`; when you specify `"credsStore": "X"`, Docker looks for a binary named `docker-credential-X`.

ecr-login/cli/docker-credential-ecr-login/main.go
```diff
@@ -14,13 +14,36 @@
...
+
 	ecr "github.com/awslabs/amazon-ecr-credential-helper/ecr-login"
 	"github.com/awslabs/amazon-ecr-credential-helper/ecr-login/api"
 	"github.com/awslabs/amazon-ecr-credential-helper/ecr-login/config"
+	"github.com/awslabs/amazon-ecr-credential-helper/ecr-login/version"
 	log "github.com/cihub/seelog"
 	"github.com/docker/docker-credential-helpers/credentials"
 )
 
+var vrsn bool
```
samuelkarp: Can you rename this to `versionFlag`?  Also, can you move it inside `main()`?

ecr-login/cli/docker-credential-ecr-login/main.go
```diff
@@ -14,13 +14,36 @@
...
+
+const BANNER = `amazon-ecr-credential-helper
+version: %s
+`
+
+func init() {
```
samuelkarp: Can you move this inside `main()`, just above line 48?  You can either inline it or keep it as a separate named function.  That will still make sure it executes first and avoid potential confusion with a separate `init()` function.

Makefile
```diff
@@ -49,18 +50,19 @@ all-variants: linux-amd64 darwin-amd64 windows-amd64
 .PHONY: linux-amd64
```
samuelkarp: (GitHub won't let me comment on the right line) Can you adjust the `$(LOCAL_BINARY)` target as well?

ecr-login/cli/docker-credential-ecr-login/main.go
```diff
@@ -14,13 +14,36 @@
...
+	flag.BoolVar(&vrsn, "version", false, "print version and exit")
```
samuelkarp: Go's `flag` package doesn't properly support GNU-style long options.  If you want to use the `flag` package, I'd rather only support the `-v` short option instead of also having a long option that is non-standard.

ecr-login/cli/docker-credential-ecr-login/main.go
```diff
@@ -14,13 +14,36 @@
...
 	"github.com/awslabs/amazon-ecr-credential-helper/ecr-login/api"
 	"github.com/awslabs/amazon-ecr-credential-helper/ecr-login/config"
+	"github.com/awslabs/amazon-ecr-credential-helper/ecr-login/version"
 	log "github.com/cihub/seelog"
 	"github.com/docker/docker-credential-helpers/credentials"
 )
 
+var vrsn bool
+
+const BANNER = `amazon-ecr-credential-helper
```
samuelkarp: Can you rename this to `banner`?  This constant doesn't need to be exported (start with a capital letter), and the Go convention of [Mixed Caps](https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps) applies to constants as well.

ecr-login/version/version.go
```diff
@@ -0,0 +1,4 @@
+package version
+
+// VERSION indicates which version of the binary is running.
+var VERSION = "master"
```
samuelkarp: Can you rename this as `Version`?

scripts/build_binary.sh
```diff
@@ -30,7 +30,9 @@ mkdir -p $1
 
 cd "${ROOT}"
 
+version_ldflags="-X github.com/awslabs/amazon-ecr-credential-helper/ecr-login/version.VERSION=${2}"
```
samuelkarp: The behavior that you have for `version.VERSION` defaulting to `"master"` isn't quite working because you're still setting `version_ldflags` when `${2}` is empty.  Can you do this instead?

```bash
version_ldflags=""
if [[ -n "${2}" ]]; then
  version_ldflags="-X github.com/awslabs/amazon-ecr-credential-helper/ecr-login/version.VERSION=${2}"
fi
```

scripts/build_binary.sh
```diff
@@ -30,7 +30,9 @@ mkdir -p $1
 
 cd "${ROOT}"
```
samuelkarp: (GitHub won't let me comment on the right line)

Can you remove all the commented-out lines starting with `# TODO!!!`?  Your new code replaces that.

ecr-login/api/factory.go
```diff
@@ -65,6 +74,7 @@ func (defaultClientFactory DefaultClientFactory) NewClient(awsSession *session.S
 
 // NewClientWithOptions Create new client with Options
 func (defaultClientFactory DefaultClientFactory) NewClientWithOptions(opts Options) Client {
+	opts.Session.Handlers.Build.PushBackNamed(userAgentHandler)
```
samuelkarp: I think you can remove it from here.  `NewClientWithOptions` is only ever called by `NewClientWithDefaults` and `NewClientFromRegion` in this program, so it'll already have the handler.  We still have this function because it can be useful in [other programs](https://github.com/aws/amazon-ecs-cli/blob/0bfcecd20eb96e5d4293b4dde9839d6ea6e5bdea/ecs-cli/modules/clients/aws/ecr/client.go#L60) which may already be setting their own user-agent.

ecr-login/cli/docker-credential-ecr-login/main.go
srizzling: I only picked flag package since its in the standard packages. Would you prefer anything else, or is supporting -v enough? Do you envision the need for more flags or arguments?

scripts/build_binary.sh
srizzling: Just realised that version will always be set, since the version is grabbed from the VERSION file. Do you want a new target for building it locally without the Version?

Also depends how much you want to follow semver, since semver supports snapshots or alplha versions between versions

srizzling:
Also as well as version, do you think it might be useful to include the commitsha of the binary as well?

ecr-login/cli/docker-credential-ecr-login/main.go
samuelkarp: I'm fine supporting just `-v` right now.  I don't really expect to have more flags or arguments, since the typical workflow for this program is to be invoked indirectly by the Docker CLI.

scripts/build_binary.sh
samuelkarp: In this particular case (and why I noticed it), the `$(LOCAL_BINARY)` target isn't currently grabbing it from the `VERSION` file.

It should typically be set if you're building with `make`, but could be unset if it's build some other way.  I think having a default value is reasonable, but it's not necessary to have a target specifically for building without the version.

> Also depends how much you want to follow semver, since semver supports snapshots or alplha versions between versions

Yeah, using a prerelease tag (like `0.0.1-dev`) is a reasonable application of semver, but I don't think we need programmatic infrastructure around that yet.

samuelkarp:
> Also as well as version, do you think it might be useful to include the commitsha of the binary as well?

Yep, that would be good too.

Also just in case you didn't see this above: Can you confirm that your contribution is licensed under the Apache 2.0 license?

srizzling:
I was going to ask about that, just so I'm clear, this isn't a CLA I'm signing. I'm just saying I accept my changes are Apache 2.0 compliant?

samuelkarp:
This is not a CLA or copyright assignment, but affirming that you are licensing your contribution under the Apache 2.0 license.

srizzling:
Ah gotcha. 

Then yes, I can confirm that my contribution is licensed under the  Apache 2.0 license. I'll get to those fixes later in the day.

samuelkarp:
Thanks for confirming!

VERSION.txt
srizzling: Fixed by c4d6021

Makefile
srizzling: Fixed by 9246903

ecr-login/cli/docker-credential-ecr-login/main.go
srizzling: Fixed by 527b0da

ecr-login/cli/docker-credential-ecr-login/main.go
srizzling: Fixed by 527b0da

ecr-login/cli/docker-credential-ecr-login/main.go
srizzling: Fixed by 527b0da

ecr-login/cli/docker-credential-ecr-login/main.go
srizzling: Fixed by 527b0da

ecr-login/version/version.go
srizzling: Fixed by 527b0da

scripts/build_binary.sh
srizzling: Fixed by 388652f

scripts/build_binary.sh
srizzling: Removed todo code in 2a7656a

Makefile
srizzling: Added $VERSION and $GITCOMMIT_SHA to both targets. Fixed in d22b66e

ecr-login/api/factory.go
srizzling: This has been fixed by 33a1bc7

srizzling:
Hi @samuelkarp, I'm sorry for the bunch of commits, but I thought it might be easier to show the fixes that way. I can squash for you as well.

> Also as well as version, do you think it might be useful to include the commitsha of the binary as well?

I have added the above feature in this change, mind you I didn't see the need to add the sha to the useragent but your opinion may change for that.

srizzling:
Also, we should add Go1.10 to the travis target, and maybe add badges for goreport?

[Go Report Card](https://goreportcard.com/report/github.com/awslabs/amazon-ecr-credential-helper)

samuelkarp:
> Also, we should add Go1.10 to the travis target, and maybe add badges for goreport?

Those both sound like good changes, want to do that in a separate PR?

samuelkarp:
@srizzling Would you mind squashing the commits?  Thanks!

srizzling:
> @srizzling Would you mind squashing the commits? Thanks!

@samuelkarp Done :) Soooo Close!

ecr-login/api/factory.go
```diff
@@ -27,17 +29,26 @@ type Options struct {
...
 	NewClientFromRegion(region string) Client
 	NewClientWithDefaults() Client
 }
+
+// DefaultClientFactory is a default implementation of the ClientFactory
 type DefaultClientFactory struct{}
 
+var userAgentHandler = request.NamedHandler{
+	Name: "ecr-login.UserAgentHandler",
+	Fn:   request.MakeAddToUserAgentHandler("amazon-ecr-credential-helper", version.Version),
```
deyangyin: Do we need to change here to be consistent with User Agent syntax?

samuelkarp: Based on how `request.MakeAddToUserAgentHandler` works, I would expect it to result in the following user-agent:

aws-sdk-go/1.12.72 (go1.10; linux; amd64) amazon-ecr-credential-helper/0.0.1

I think this is fine.

deyangyin:
> Looks good to me, thanks for contributing this! I'd like another engineer on our side to look at it before merging, but we really appreciate the effort here.

@samuelkarp Looks good to me. @srizzling Thanks for your contribution!

samuelkarp:
Merged!  Thank you for your contribution!

PR #121, opened 2018-10-28
Initial Amazon Linux packaging

samuelkarp:
*Description of changes:*
Initial packaging of an RPM for Amazon Linux 2.

Tested with `make docker-rpm`, which produced `amazon-ecr-credential-helper-0.1.0-1.amzn2.x86_64.rpm` and was install-able into an `amazonlinux:2` container.

Still TODO:
- [x] Fix git SHA embedding with `make docker-rpm` (depends on https://github.com/awslabs/amazon-ecr-credential-helper/pull/126)
- [x] Add changelog
- [x] Declare bundled dependencies
- [x] Test with actual Amazon Linux build system

/cc @jahkeup 

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,16 @@
 GITCOMMIT_SHA
 *.tar.gz
 *.tar
+
+
+# Amazon Linux RPM
+*.swp
+/BUILDROOT/
+/BUILD/
+/RPMS/
+/SRPMS/
+/SOURCES
+/SPECS/
+/.rpm-done
+/sources.tgz
+/*.rpm
--- a/Makefile
+++ b/Makefile
@@ -26,6 +26,8 @@ LINUX_AMD64_BINARY=bin/linux-amd64/$(BINARY_NAME)
 DARWIN_AMD64_BINARY=bin/darwin-amd64/$(BINARY_NAME)
 WINDOWS_AMD64_BINARY=bin/windows-amd64/$(BINARY_NAME).exe
 
+include Makefile.amazonlinux
+
 .PHONY: docker
 docker: Dockerfile
 	mkdir -p bin
@@ -96,7 +98,7 @@ get-deps:
 	go get golang.org/x/tools/cmd/goimports
 
 .PHONY: clean
-clean:
+clean: .clean-amazonlinux
 	- rm -rf ./bin
 	- rm -f GITCOMMIT_SHA
 	- rm -f release.tar.gz
--- a/Makefile.amazonlinux
+++ b/Makefile.amazonlinux
@@ -1,0 +1,48 @@
+# Copyright 2018 Amazon.com, Inc. or its affiliates. All Rights Reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License"). You
+# may not use this file except in compliance with the License. A copy of
+# the License is located at
+#
+# 	http://aws.amazon.com/apache2.0/
+#
+# or in the "license" file accompanying this file. This file is
+# distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF
+# ANY KIND, either express or implied. See the License for the specific
+# language governing permissions and limitations under the License.
+
+# Hook to perform preparation steps prior to the sources target.
+prepare-sources::
+
+sources: prepare-sources release.tar.gz
+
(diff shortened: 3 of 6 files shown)


scripts/gopath.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
```
jahkeup: I would prefer to see this setup and build harness moved into the spec if this code isn't reused elsewhere (in this PR it isn't, but maybe you have another use case on the way).

To my eyes, this external file causes things to be a bit more complicated than need be and in effect obfuscates the simple mechanism at heart of it: setting up the directory for build. The spec file can gain the necessary environment variables, placement of the project (for GOPATH building), and the build & pluck the built binaries out of the locations the project was already building them into during `%install`.

scripts/gopath.sh
```diff
@@ -0,0 +1,30 @@
...
+export GOPATH="${BUILDDIR}"
+export SRCPATH="${BUILDDIR}/src/github.com/awslabs/amazon-ecr-credential-helper"
+declare -a LINKS=("ecr-login" "scripts" "Makefile" "VERSION" ".git")
+mkdir -p "${SRCPATH}"
+for LINK in "${LINKS[@]}"; do
+	ln -s "${TOPWD}/${LINK}" "${SRCPATH}"
+done
+cd "${SRCPATH}"
+ls
+source "${SRCPATH}/scripts/shared_env"
```
jahkeup: Is this still necessary? It looks to me like the variables there are not appropriate for the version of Go used nor for the usage of tool on the path for build.

scripts/gopath.sh
samuelkarp: Done

scripts/gopath.sh
samuelkarp: Done

samuelkarp:
@jahkeup This is pretty much done, but I'm leaving it as WIP because I need your help testing this in the Amazon Linux build system.

packaging/amazon-linux/amazon-ecr-credential-helper.spec
```diff
@@ -0,0 +1,126 @@
...
+Provides:       bundled(golang(golang.org/x/sys/windows))
+
+%description
+The Amazon ECR Docker Credential Helper is a credential helper for the Docker
+daemon that makes it easier to use Amazon Elastic Container Registry.
+
+%prep
+%setup -c
+
+%build
```
jahkeup: One last update: the `%build` section can be simplified a bit:

```rpm-spec
%build
export GOPATH="$(pwd)/_gopath"
mkdir -p "_gopath/src/github.com/awslabs"
ln -sv "$(pwd)" "_gopath/src/github.com/awslabs/amazon-ecr-credential-helper"
cd "_gopath/src/github.com/awslabs/amazon-ecr-credential-helper"
make
```

packaging/amazon-linux/amazon-ecr-credential-helper.spec
```diff
@@ -0,0 +1,126 @@
...
+cd "${SRCPATH}"
+ls
+export BIN_DIRECTORY="${TOPWD}"
+make
+
+%install
+rm -rf %{buildroot}
+mkdir -p %{buildroot}/%{_bindir}
+install -p -m 0755 \
+  local/docker-credential-ecr-login \
```
jahkeup: And then if you do the above, this would need to be updated to install from `bin/local/docker-credential-ecr-login`

packaging/amazon-linux/amazon-ecr-credential-helper.spec
samuelkarp: Done

packaging/amazon-linux/amazon-ecr-credential-helper.spec
samuelkarp: Done

samuelkarp:
This works, but I'm waiting for https://github.com/awslabs/amazon-ecr-credential-helper/pull/126 and then I'll update this to remove the `.git` directory from the SRPM.

samuelkarp:
@jahkeup I'd appreciate your help in testing this again; I've made some modifications to the spec file after rebasing on #126 and would like to validate that it still works properly.

PR #122, opened 2018-10-30
deps: replace seelog with logrus

samuelkarp:
*Description of changes:*
Replace seelog with logrus

logrus better aligns with [ecs-cli](https://github.com/aws/amazon-ecs-cli) and allows for structured logging.  logrus also has a relatively straightforward mechanism to override the output, so a different backing logger can be used.

Notable changes:
* Log lines are formatted differently, using the default logrus logger instead of the custom seelog logger.
* logrus does not rotate logs.  There is potential for unbounded growth of logs, but in practice docker-credential-ecr-login should not be very verbose.


ecr-login/api/client.go
```diff
@@ -76,7 +76,11 @@ func (self *defaultClient) GetCredentials(serverURL string) (*Auth, error) {
 	if err != nil {
 		return nil, err
 	}
-	log.Debugf("Retrieving credentials for %s in %s (%s)", registry.ID, registry.Region, serverURL)
+	logrus.
+		WithField("registry", registry.ID).
```
PettitWesley: Should line 79 and 80 be one line? This looks weird to my eyes

ecr-login/ecr.go
```diff
@@ -42,31 +42,34 @@ func (ECRHelper) Delete(serverURL string) error {
 }
 
 func (self ECRHelper) Get(serverURL string) (string, string, error) {
-	defer log.Flush()
+	//defer log.Flush()
```
PettitWesley: Should this commented out code be removed?

PettitWesley:
General nit/note: on the CLI team, we always split the vendor changes out into a separate commit. I'm not sure what the usual practice is for this project, but I think splitting the changes that way makes it nicer to read/review.

samuelkarp:
@PettitWesley Fixed

ecr-login/api/client.go
nmeyerhans: I much prefer it broken across multiple lines like this, but why not use [WithFields](https://godoc.org/github.com/sirupsen/logrus#Logger.WithFields)?

ecr-login/cache/build.go
```diff
@@ -22,13 +22,13 @@ import (
...
-	log "github.com/cihub/seelog"
-	homedir "github.com/mitchellh/go-homedir"
+	"github.com/mitchellh/go-homedir"
+	"github.com/sirupsen/logrus"
 )
 
 func BuildCredentialsCache(awsSession *session.Session, region string, cacheDir string) CredentialsCache {
 	if os.Getenv("AWS_ECR_DISABLE_CACHE") != "" {
-		log.Debug("Cache disabled due to AWS_ECR_DISABLE_CACHE")
+		logrus.Debug("Cache disabled due to AWS_ECR_DISABLE_CACHE")
```
jahkeup: I'd argue this is a warning if this isn't intended to be used normally.

ecr-login/cache/build.go
```diff
@@ -39,17 +39,15 @@ func BuildCredentialsCache(awsSession *session.Session, region string, cacheDir
 
 	cacheDir, err := homedir.Expand(cacheDir)
 	if err != nil {
-		log.Debugf("Could not expand cache path: %s", err)
-		log.Debug("Disabling cache")
+		logrus.WithError(err).Debug("Could not expand cache path, disabling cache")
```
jahkeup: Neat! That's a handy method.

ecr-login/config/seelog.go
```diff
@@ -15,46 +15,32 @@ package config
...
-				<console />
-			</filter>
-		</outputs>
-		<formats>
-			<format id="main" format="%UTCDate(2006-01-02T15:04:05Z07:00) [%LEVEL] %Msg%n" />
-		</formats>
-	</seelog>
-`
-	return config
+	file, err := os.OpenFile(logfile, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0666)
```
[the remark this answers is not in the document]
jahkeup: umask should be respected per `os.OpenFile`'s [godoc](https://godoc.org/os#OpenFile):

> OpenFile is the generalized open call; most users will use Open or Create instead. It opens the named file with specified flag (O_RDONLY etc.) and perm (before umask), if applicable. If successful, methods on the returned File can be used for I/O. If there is an error, it will be of type *PathError.

+1 for 664

ecr-login/config/seelog.go
```diff
@@ -15,46 +15,32 @@ package config
```
jahkeup: Can we rename this file to reflect the updated code?

ecr-login/api/client.go
samuelkarp:
```go
logrus.
	WithFields(Fields{
		"registry": registry.ID,
		"region": registry.Region,
```

This looks like it adds an extra line and extra indentation.

ecr-login/api/client.go
```diff
@@ -145,9 +152,9 @@ func (self *defaultClient) getAuthorizationToken(registryID string) (*Auth, erro
 	if err != nil || output == nil {
 		if err == nil {
 			if registryID == "" {
-				err = fmt.Errorf("Mising AuthorizationData in ECR response for default registry")
+				err = fmt.Errorf("ecr: Mising AuthorizationData in ECR response for default registry")
```
jahkeup: Can we simply wrap `err` with `"ecr: "` (or `errors.Wrap(err, "ecr")` if you have it here) when its returned by this func to make these errors consistent?

ecr-login/api/client.go
jahkeup: I prefer the `WithFields` version, but don't have an issue with `WithField`.

nmeyerhans: I'd remove the line break after the module name and write it as
```
logrus.WithFields(Fields{
 ...
```
But I won't object if you want to stick with the chained `WithField()` form.

ecr-login/cache/build.go
samuelkarp: It's an intentional option.  I don't think this should be a warning.

ecr-login/config/seelog.go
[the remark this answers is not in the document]
samuelkarp: 0666 was used by seelog.  I can change to 0664.

ecr-login/config/seelog.go
samuelkarp: Yes

ecr-login/api/client.go
samuelkarp: `err` is nil here.  This is sanity-checking that `output` is not nil.

ecr-login/config/seelog.go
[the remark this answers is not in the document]
samuelkarp: Done.

ecr-login/config/seelog.go
samuelkarp: Done

ecr-login/api/client.go
jahkeup: Yeah, I see these are used when `err == nil`, however there is another error bubbled up when `err != nil`, which I think should also be prefixed with `ecr: ` as any error from this point is emitted/handled in the context of `ecr: `.

samuelkarp: `github.com/pkg/errors` was not already a dependency, but I generally like that package so I've added it and used `errors.Wrap`.

ecr-login/api/client.go
```diff
@@ -145,12 +154,12 @@ func (self *defaultClient) getAuthorizationToken(registryID string) (*Auth, erro
...
 			} else {
-				err = fmt.Errorf("Missing AuthorizationData in ECR response for %s", registryID)
+				err = fmt.Errorf("ecr: Missing AuthorizationData in ECR response for %s", registryID)
 			}
 		}
-		return nil, err
+		return nil, errors.Wrap(err, "ecr: Failed to get authorization token")
```
jahkeup: With this wrapped this way, I think the `ecr: ` prefix should be dropped from the `fmt.Errorf`'s above so we don't get:

`ecr: Failed to get authorization token: ecr: Missing AuthorizationData in ECR response for ...`

ecr-login/api/client.go
```diff
@@ -145,12 +154,12 @@ func (self *defaultClient) getAuthorizationToken(registryID string) (*Auth, erro
 	if err != nil || output == nil {
 		if err == nil {
 			if registryID == "" {
-				err = fmt.Errorf("Mising AuthorizationData in ECR response for default registry")
+				err = fmt.Errorf("ecr: Mising AuthorizationData in ECR response for default registry")
```
jahkeup: nit: `s/Mising/Missing/`

samuelkarp: Done

ecr-login/api/client.go
samuelkarp: Done

ecr-login/api/client.go
```diff
@@ -85,10 +91,13 @@ func (self *defaultClient) GetCredentialsByRegistryID(registryID string) (*Auth,
 	cachedEntry := self.credentialCache.Get(registryID)
 	if cachedEntry != nil {
 		if cachedEntry.IsValid(time.Now()) {
-			log.Debugf("Using cached token for %s", registryID)
+			logrus.WithField("registry", registryID).Debug("Using cached token for")
```
PettitWesley: nit: probably the message would read fine just as "Using cached token"

samuelkarp: Fixed

PR #126, opened 2018-11-07
makefile: targets to create source archives

samuelkarp:
release-tarball: creates a source archive with `git archive`
release-tarball-no-vendor: same as release-tarball, but excludes the
vendor folder

Both archives have an embedded GITCOMMIT_SHA file which records the commit from which the archive was created.  This commit is then embedded into the resulting binary after building.


Makefile
```diff
@@ -65,6 +66,24 @@ $(WINDOWS_AMD64_BINARY): $(SOURCES)
 	./scripts/build_variant.sh windows amd64 $(VERSION) $(GITCOMMIT_SHA)
 	@mv ./bin/windows-amd64/$(BINARY_NAME) ./$(WINDOWS_AMD64_BINARY)
 
+GITCOMMIT_SHA: $(GITFILES)
+	git rev-parse --short HEAD > GITCOMMIT_SHA
```
jahkeup: It would be wise to specify a length to `--short` so that no git or distribution modifications cause this to change length.

samuelkarp: I can if you'd like; I don't have a strong opinion here.  I'm not using this SHA for anything other than an indication as to the source from which the binary was built (debugging); there is no programmatic dependency on it or its length right now.

Any particular suggestion on the length?

jahkeup: 7 :+1: 	

I don't see any distributed binaries so there's not really a notion of a "known length". 7 is the default according to `man git-rev-parse`, so that seems like a sane constant to provide. Additionally, Github looks to be using 7 char SHAs when there are short SHA links.

Makefile
```diff
@@ -18,6 +18,7 @@ all: build
 SOURCEDIR=./ecr-login
 SOURCES := $(shell find $(SOURCEDIR) -name '*.go')
 VERSION := $(shell cat VERSION)
+GITFILES := $(shell find ".git/")
 GITCOMMIT_SHA := $(shell git rev-parse --short HEAD)
```
jahkeup: Is there any reason to or not to eliminate this variable? On the surface, this look like it could be replaced in many usages with a dependency on the GITCOMMIT_SHA file and `$(shell cat GITCOMMIT_SHA)`.

Not a blocker, but I'm curious.

samuelkarp: Oh, I did mean to remove the GITCOMMIT_SHA variable.  Will do that.

PR #136, opened 2018-12-25
Debian

samuelkarp:
*Description of changes:*

Initial Debian packaging for the Amazon ECR Credential Helper.  This closes [ITP #916982](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=916982) once it gets merged, signed, and uploaded.

@nmeyerhans, I would appreciate your review here.

This is broken up into two commits:
1. This first commit just removes the `vendor` folder.  Debian policy prefers the dependencies to be packaged separately, and we use those separately-packaged dependencies.
2. The second commit is the actual Debian packaging, including a script `debian/docker-deb` that I use to produce all the proper artifacts.


samuelkarp:
Note: I expect pretty much all of the Travis runs to fail, since the dependencies are no longer vendored here.  Easiest way to address this is probably by setting a [blocklist](https://docs.travis-ci.com/user/customizing-the-build/#building-specific-branches) and prevent Travis from doing a build, but a (perhaps) better approach would be to make Travis perform a real Debian build.

samuelkarp:
I intend to add some documentation similar to [this](https://github.com/awslabs/amazon-ecr-credential-helper/blob/ba1506f8376b34bd405b98a3ab7b90b8b6933d22/docs/packaging-amazon-linux.md) (though hopefully without the broken markdown...).

debian/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM debian:sid
+
+RUN DEBIAN_FRONTEND=noninteractive apt-get -o Acquire::ForceIPv4=true update && \
+    apt-get -o Acquire::ForceIPv4=true install -yq \
```
nmeyerhans: You need to set `DEBIAN_FRONTEND` for the "install" line, too.

debian/rules
```diff
@@ -0,0 +1,32 @@
+#!/usr/bin/make -f
+
+export DH_VERBOSE=1
+
+export DH_OPTIONS
+
+export DH_GOPKG := github.com/awslabs/amazon-ecr-credential-helper
+
+VERSION := $(shell cat VERSION)
+GITCOMMIT_SHA := $(shell git rev-parse --short HEAD)
```
nmeyerhans: If you really want to do this, you'll need to take extra steps to prevent `dpkg-source` from pruning your .git directory. As it is, if you try to run a build based on the debianized sources generated by the `docker-deb` script, you'll see:

```
debian/rules binary
fatal: not a git repository (or any of the parent directories): .git
```
(Note that it's not  actually fatal to the build, but GITCOMMIT_SHA is set to the empty string.)

See the `--diff-ignore` and `--extend-diff-ignore` flags to [dpkg-source](https://manpages.debian.org/stretch/dpkg-dev/dpkg-source.1.en.html) and note that `dpkg-buildpackage` will invoke it as `dpkg-source -i -I -b .` by default.

And if you do this, don't forget to add a lintian override to suppress the warning that it'd otherwise generate.

debian/control
```diff
@@ -0,0 +1,30 @@
...
+XS-Go-Import-Path: github.com/awslabs/amazon-ecr-credential-helper
+
+Package: amazon-ecr-credential-helper
+Architecture: any
+Depends:
+ ${misc:Depends},
+ ${shlibs:Depends}
+Built-Using: ${misc:Built-Using}
+Description: Amazon ECR Credential Helper for Docker
+ A Docker credential helper to automatically manage credentials for Amazon ECR
```
nmeyerhans: It's probably a good idea to extend the long description somewhat. If this package shows up in some search results, how will I know that it solves the problem I'm having? What does it mean to "automatically manage credentials"?

debian/rules
samuelkarp: Thanks, I thought I had changed this.  There should be a `GITCOMMIT_SHA` file in the orig tarball, so I should be able to do `GITCOMMIT_SHA := $(shell cat GITCOMMIT_SHA)`.  I will ensure that works and update.

samuelkarp: Fixed.

debian/Dockerfile
samuelkarp: Fixed.

debian/control
samuelkarp: Done.

samuelkarp:
@nmeyerhans I've addressed your comments and [uploaded a package](https://mentors.debian.net/package/amazon-ecr-credential-helper).

debian/control
```diff
@@ -0,0 +1,42 @@
...
+ golang-github-pkg-errors-dev
+Standards-Version: 4.3.0
+Homepage: https://github.com/awslabs/amazon-ecr-credential-helper
+Vcs-Browser: https://github.com/awslabs/amazon-ecr-credential-helper
+Vcs-Git: https://github.com/awslabs/amazon-ecr-credential-helper
+XS-Go-Import-Path: github.com/awslabs/amazon-ecr-credential-helper
+
+Package: amazon-ecr-credential-helper
+Architecture: any
+Depends:
```
nmeyerhans: Should this package declare a dependency on docker.io? Is it actually useful without it?

samuelkarp: I used `Enhances: docker.io` to capture the relationship.  This package does not depend on Docker at runtime, but is designed to work with it.

The Docker project builds packages for Debian as well: https://docs.docker.com/install/linux/docker-ce/debian/.  Those packages are called `docker-ce` instead of `docker.io`.  I'd like this package to work with both the Debian project's `docker.io` package and the Docker project's `docker-ce` package.

nmeyerhans: Enhances and Depends have different meanings. In particular, Enhances is a fairly weak relationship. Depends is much stronger. With Enhances, it's possible to have the credential helper installed with no Docker package at all. Is that a useful configuration? If so, then enhances may be reasonable, though possibly also with a Recommends or Suggests relationship on Docker packages. If not, then Depends is needed.

You can declare a dependency relationship that can be met by the various different docker packages with something like `Depends: docker.io | docker-ce`. `docker.io` should be listed first, to ensure that, if a Docker package is installed in order to fulfill the dependency, it will be the one from the Debian archive.

See [ch 7 of debian-policy](https://www.debian.org/doc/debian-policy/ch-relationships.html) for details, and in particular [section 7.2](https://www.debian.org/doc/debian-policy/ch-relationships.html#binary-dependencies-depends-recommends-suggests-enhances-pre-depends)

samuelkarp: I really don't think a Depends relationship is warranted here.  The application does not depend on any facility that Docker provides at runtime, and I can conceive of situations where a user might want to install amazon-ecr-credential-helper without Docker (e.g., if they wanted to use amazon-ecr-credential-helper with another Docker registry-compatible container runtime like cri-o, containerd, rkt, or with a utility like [img](https://github.com/genuinetools/img)).

I think Enhances makes sense given that amazon-ecr-credential-helper provides additional usability for Docker, but I think a Recommends or Suggests could be fine as well.

Thanks for telling me about the use of the `|` operator in dependency relationships; I didn't know that was possible.

Should I change Enhances to `Recommends: docker.io | docker-ce`?

samuelkarp: Per our discussion this morning I've added a `Recommends: docker.io | docker-ce` and changed the Enahnces to `Enhances: docker.io, docker-ce`.

samuelkarp:
@nmeyerhans found that he couldn't rebuild this package: https://paste.debian.net/1061058/.  It appears that the golang-github-aws-aws-sdk-go-dev package changed in sid, and the mocks that I generated are not compatible with the updated interface in the package.  I've opened https://github.com/awslabs/amazon-ecr-credential-helper/issues/145 to track the issue in the main project.

PR #138, opened 2019-01-02
snap: initial packaging

samuelkarp:
*Description of changes:*
Initial packaging as a [snap](https://snapcraft.io/) package.

This needs review in the following areas:

* Is it possible for us to get away from needing `classic` confinement?  The `home` plug does not allow reading hidden directories, and we need to read `~/.aws` (and write to `~/.ecr`, but that could really be confined inside the snap's `$HOME` instead of the user's).
* The `snapcore/snapcraft` Docker image is based on Ubuntu 16.04, which does not have a new enough `golang-go` package for this project to build.  Consequently, I wrote a Dockerfile to build from Ubuntu 18.04 which has Go 1.10 available.
* The snap name is `docker-credential-ecr-login` as that seems to be the only way for a binary named `docker-credential-ecr-login` to be added to the user's `$PATH`, and Docker requires that naming scheme.  Is it possible to name the snap something more meaningful (like `amazon-ecr-credential-helper` while still exposing a binary that appears as `docker-credential-ecr-login` in `$PATH`?
* I used `override-build` to control the build process and actually use our `Makefile`, but it does then mean manually handling copying files to the appropriate location.  Is there a way to either still use our `Makefile` or to inject the appropriate Go options (including injecting the project version and commit into the binary)?

/cc @davdunc

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,10 @@
 GITCOMMIT_SHA
 *.tar.gz
 *.tar
+
+# Snaps
+*.snap
+parts/
+prime/
+stage/
+snap/.snapcraft
--- a/scripts/snap-in-docker.sh
+++ b/scripts/snap-in-docker.sh
@@ -1,0 +1,22 @@
+#!/bin/bash
+
+# Normalize to working directory being build root (up one level from ./snap)
+SOURCES=$( cd "$( dirname "${BASH_SOURCE[0]}" )/.." && pwd )
+
+# docker build needs context, but we don't really want to provide any
+TMPCONTEXT=$(mktemp -d)
+docker build \
+       --network=host \
+       -t amazon/amazon-ecr-credential-helper:snapcraft \
+       -f scripts/snapcraft.dockerfile \
+       "${TMPCONTEXT}"
+
+rmdir "${TMPCONTEXT}"
+
+docker run \
+       --rm -it\
+       --net=host \
+       -v ${SOURCES}:/build \
+       amazon/amazon-ecr-credential-helper:snapcraft \
+       snapcraft \
+       $@
--- a/scripts/snapcraft.dockerfile
+++ b/scripts/snapcraft.dockerfile
@@ -1,0 +1,11 @@
+FROM ubuntu:18.04
+
+RUN DEBIAN_FRONTEND=noninteractive apt-get update \
+ && DEBIAN_FRONTEND=noninteractive apt-get install -y \
+    golang-go \
+    make \
+    git \
+    snapcraft
+
+WORKDIR /build
+CMD ["snapcraft"]
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -1,0 +1,34 @@
+name: docker-credential-ecr-login
+version: '0.2.0-dev'
+summary: Amazon ECR Credential Helper for Docker
+description: |
+  A Docker helper to automatically manage credentials for Amazon ECR.
+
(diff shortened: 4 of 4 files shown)


samuelkarp:
Opened https://github.com/awslabs/amazon-ecr-credential-helper/issues/139 for failing CI, which is unrelated to the changes on this branch.

stilvoid:
Hi @samuelkarp, I'm looking into this. To address your questions:

* Yes, we should be able to do this without `classic` by using the `[personal-files](https://docs.snapcraft.io/the-personal-files-interface/9357)` interface.

* I've got an updated recipe using `core18` that seems to be working ok

* It's possible to ask for an auto-alias for a snap package but for this one I'm not sure it's worth it. `snap search` checks description text as well as the package name. It's possible though if you think it's worth fighting for :)

* Removing the need for a complex `override-build` is what I'm looking at right now. I can get the package to build ok without using the makefile at all so I might just be able to get away with setting a few build-time environment variables. I'll let you know when I make progress.

My branch is here: <https://github.com/stilvoid/amazon-ecr-credential-helper/tree/snap/snap>

samuelkarp:
Hi @stilvoid, thank you for taking a look!

> Yes, we should be able to do this without classic by using the personal-files interface.

Is `personal-files` new?  I hadn't seen it before and it addresses one of my main complaints with snaps. :smile:  I think it'll work perfectly for this.

> I've got an updated recipe using core18 that seems to be working ok

Does `core18` require building or running on Ubuntu 18.04?  16.04 is still an LTS until 2021.

> It's possible to ask for an auto-alias for a snap package but for this one I'm not sure it's worth it. snap search checks description text as well as the package name. It's possible though if you think it's worth fighting for :)

I'd like to have the auto-alias.  I was talking with @davdunc about this (as he helps manage the Snap store account for AWS) and he did end up creating the snap listing for the name we want, but it ended up falling off both of our radars for now.  I'll restart that discussion.

> Removing the need for a complex override-build is what I'm looking at right now. I can get the package to build ok without using the makefile at all so I might just be able to get away with setting a few build-time environment variables. I'll let you know when I make progress.

That might work, though I still want to make sure that the snap includes the manual page and the licensing files.

stilvoid:
> Does `core18` require building or running on Ubuntu 18.04? 16.04 is still an LTS until 2021.

I'm building and running from Arch Linux so... nope ;)

>I still want to make sure that the snap includes the manual page and the licensing files.
Can snaps even do that? The man page wouldn't be able to live under `/usr/share/man` so I'm not sure how `man` would index them. Have you seen an example of a package that does this?

samuelkarp:
> Can snaps even do that? The man page wouldn't be able to live under /usr/share/man so I'm not sure how man would index them. Have you seen an example of a package that does this?

I'm hoping that they will in the future :)

I'm really more concerned about the licensing files (LICENSE, NOTICE, and THIRD-PARTY-LICENSES) than I am about the man page, though I would like the man page to be included as well.

samuelkarp:
> So we can certainly include those in the snap but they'll just live in /var/lib/snapd/snap/whatever so I'm not certain how useful it would be.

That's fine, I'd rather have them included in the package than not included.

Want to send me a PR for the `confinement`, `personal-files`, and `license` changes?  Otherwise I can update this branch with a new commit and credit you as the author.

samuelkarp:
Looks like https://github.com/docker-snap/docker-snap/issues/10 covers some complications that would make it hard to use both the Docker snap and a snap of this credential helper together.

PR #149, opened 2019-01-29
docs: add Debian and Ubuntu packaging and install

samuelkarp:
docs: add Debian and Ubuntu packaging and install

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,37 @@ Once you have installed the credential helper, see the
 [Configuration section](#Configuration) for instructions on how to configure
 Docker to work with the helper.
 
+### Debian Buster (and future versions)
+You can install the Amazon ECR Credential Helper from the Debian Buster
+archives.  This package will also be included in future releases of Debian.
+
+[![Debian Testing package](https://repology.org/badge/version-for-repo/debian_testing/amazon-ecr-credential-helper.svg)](https://repology.org/metapackage/amazon-ecr-credential-helper/versions)
+[![Debian Unstable package](https://repology.org/badge/version-for-repo/debian_unstable/amazon-ecr-credential-helper.svg)](https://repology.org/metapackage/amazon-ecr-credential-helper/versions)
+
+```bash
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -57,6 +57,37 @@ Once you have installed the credential helper, see the
 [Configuration section](#Configuration) for instructions on how to configure
 Docker to work with the helper.
 
+### Debian Buster (testing) and Debian Sid (unstable)
```
nmeyerhans: I'd try to take a more general approach here. Buster won't be testing forever, and there will be a new testing at some point. This text is thus subject to bitrot. Might be worth instead describing that this package will be included in buster and future releases without assuming specific name mappings.

README.md
```diff
@@ -57,6 +57,37 @@ Once you have installed the credential helper, see the
...
+```bash
+$ sudo apt update
+$ sudo apt install amazon-ecr-credential-helper
+```
+
+Once you have installed the credential helper, see the
+[Configuration section](#Configuration) for instructions on how to configure
+Docker to work with the helper.
+
+### Ubuntu 19.04 Disco Dingo
```
nmeyerhans: Similar to Debian, I'd try to pick language that'll continue to be meaningful as future releases go. We can expect the package to be installable on 19.10, but this language will be broken. Maybe "19.04 and newer" or "19.04+" or something like that.

README.md
samuelkarp: My intent was just to update this when Buster becomes stable.  Do you think that's too complicated?

README.md
samuelkarp: As above

README.md
nmeyerhans: Yeah, seems annoying to have to make commits to this repo just to note activity in third-party projects.

samuelkarp: Done

README.md
samuelkarp: Done

PR #167, opened 2019-07-03
Extra notes for installing from source

Torniojaws:
*Issue #, if available:* 101

Added note about golang PATH definitions being required prior to installing from source. Otherwise the install will fail silently.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,23 @@ Docker to work with the helper.
 To build and install the Amazon ECR Docker Credential Helper, we suggest Go
 1.9+, `git` and `make` installed on your system.
 
+If you just installed Go, make sure you also have added it to your PATH or 
+Environment Vars (Windows). For example:
+
+```
+$ export GOPATH=$HOME/gocode
+$ export PATH=$PATH:$GOPATH/bin
+```
+
+Or in Windows:
+
+```
+<your existing PATH definitions>;C:\Users\<username>\gocode\bin\
+```
+
+If you haven't defined the PATH, the command below will fail silently, and
+running `docker-credential-ecr-login` will output: `command not found`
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -107,6 +107,23 @@ Docker to work with the helper.
 To build and install the Amazon ECR Docker Credential Helper, we suggest Go
 1.9+, `git` and `make` installed on your system.
 
+If you just installed Go, make sure you also have added it to your PATH or 
+Environment Vars (Windows). For example:
+
+```
+$ export GOPATH=$HOME/gocode
```
samuelkarp: nit: In recent versions of Go (1.10 or greater), the GOPATH is defaulted to `~/go`.  I'd prefer that we recommend `$HOME/go` or `~/go` to stay consistent.

README.md
```diff
@@ -107,6 +107,23 @@ Docker to work with the helper.
...
+$ export PATH=$PATH:$GOPATH/bin
+```
+
+Or in Windows:
+
+```
+<your existing PATH definitions>;C:\Users\<username>\gocode\bin\
```
samuelkarp: Same as above, this could be `%USERPROFILE%\go\bin`.  And I think you might want to add `setx GOPATH %USERPROFILE%\go` to be consistent with the advice for Linux.

(Windows commands are documented in [the Go wiki](https://github.com/golang/go/wiki/SettingGOPATH))

README.md
Torniojaws: Sounds good!

Torniojaws: Fixed.

README.md
Torniojaws: Fixed.

Issue #162, opened 2019-04-24
No logs are written unless ~/.ecr/log is created manually


jbergknoff-rival:
I recently installed ECR credential helper from source (`make docker`) on a new machine and noticed that it's not writing logs. There is ~/.ecr/cache.json but no ~/.ecr/log directory. If I make the ~/.ecr/log directory then logs will be written there.

```
$ docker-credential-ecr-login -v
amazon-ecr-credential-helper
Version:    0.3.0
Git commit: 10153a9
```

samuelkarp:
Thank you for the bug report!  I'm able to reproduce this:

<details>

```
skarp@laptop:~$ ls ~/.ecr
cache.json
skarp@laptop:~$ docker-credential-ecr-login get <<< not-valid.example.com
ERRO[0000] Error parsing the serverURL                   error="docker-credential-ecr-login can only be used with Amazon Elastic Container Registry." serverURL=not-valid.example.com
credentials not found in native keychain
skarp@laptop:~$ mkdir ~/.ecr/log
skarp@laptop:~$ docker-credential-ecr-login get <<< not-valid.example.com
credentials not found in native keychain
skarp@laptop:~$ cat ~/.ecr/log/ecr-login.log 
time="2019-04-24T13:01:39-07:00" level=error msg="Error parsing the serverURL" error="docker-credential-ecr-login can only be used with Amazon Elastic Container Registry." serverURL=not-valid.example.com
```

</details>

PR #168, opened 2019-07-03
log: create log directory if missing

samuelkarp:
*Issue #, if available:*
https://github.com/awslabs/amazon-ecr-credential-helper/issues/162

*Description of changes:*
Ensure the log directory is created so that logs are put in the log file instead of standard out.

cc @jbergknoff-rival, @sipsma

### DIFF
--- a/ecr-login/config/log.go
+++ b/ecr-login/config/log.go
@@ -27,14 +27,19 @@ func SetupLogger() {
 }
 
 func logrusConfig() {
-	logfile, err := homedir.Expand(GetCacheDir() + "/log/ecr-login.log")
+	logdir, err := homedir.Expand(GetCacheDir() + "/log")
 	if err != nil {
-		fmt.Errorf("%v", err)
-		logfile = "/tmp/.ecr/log/ecr-login.log"
+		fmt.Fprintf(os.Stderr, "log: failed to find directory: %v", err)
+		logdir = os.TempDir()
 	}
 	// Clean the path to replace with OS-specific separators
-	logfile = filepath.Clean(logfile)
-	file, err := os.OpenFile(logfile, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0664)
+	logdir = filepath.Clean(logdir)
+	err = os.MkdirAll(logdir, os.ModeDir|0700)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "log: failed to create directory: %v", err)
+		logdir = os.TempDir()
+	}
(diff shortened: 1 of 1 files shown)


ecr-login/config/log.go
```diff
@@ -27,14 +27,19 @@ func SetupLogger() {
 }
 
 func logrusConfig() {
-	logfile, err := homedir.Expand(GetCacheDir() + "/log/ecr-login.log")
+	logdir, err := homedir.Expand(GetCacheDir() + "/log")
 	if err != nil {
-		fmt.Errorf("%v", err)
-		logfile = "/tmp/.ecr/log/ecr-login.log"
+		fmt.Errorf("log: failed to find directory: %v", err)
```
mxpv: The result of `fmt.Errorf` not used?

samuelkarp: ...thank you for catching my brain fart.  This should be `fmt.Sprintf(os.Stderr, ...)`.  Fix incoming...

samuelkarp: Done

samuelkarp:
This is the expected output after this fix:

```
skarp@laptop:~$ ls ~/.ecr
cache.json
skarp@laptop:~$ docker-credential-ecr-login get <<< not-valid.example.com
credentials not found in native keychain
skarp@laptop:~$ cat ~/.ecr/log/ecr-login.log 
time="2019-07-03T16:14:34-07:00" level=error msg="Error parsing the serverURL" error="docker-credential-ecr-login can only be used with Amazon Elastic Container Registry." serverURL=not-valid.example.com
```

ecr-login/config/log.go
```diff
@@ -27,14 +27,19 @@ func SetupLogger() {
...
 	// Clean the path to replace with OS-specific separators
-	logfile = filepath.Clean(logfile)
-	file, err := os.OpenFile(logfile, os.O_RDWR|os.O_CREATE|os.O_APPEND, 0664)
+	logdir = filepath.Clean(logdir)
+	err = os.MkdirAll(logdir, os.ModeDir|0700)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "log: failed to create directory: %v", err)
+		logdir = os.TempDir()
+	}
+	file, err := os.OpenFile(filepath.Join(logdir, "ecr-login.log"), os.O_RDWR|os.O_CREATE|os.O_APPEND, 0664)
```
mxpv: I think this also needs to be closed eventually.
Something like `logrus.RegisterExitHandler(...)` should do the job.

samuelkarp: I'll add that, but I don't think it's strictly necessary to `close(2)` a file when the process exits.

Issue #184, opened 2019-09-19
`make docker` fails when the repo is cloned as a submodule


scjudd:
If the repo is cloned as a submodule, `make docker` fails. This is due to the fact that `GITCOMMIT_SHA` doesn't exist before `make` runs within the build container, and the true `.git/` directory is outside of the Docker build context when we're in a submodule.

I was able to fix this by making `GITCOMMIT_SHA` a dependency of the `docker` make target in the `Makefile`, such that `GITCOMMIT_SHA` already exists in the Docker build context before `make` runs. I have opened #185 with the fix.

### Example
```console
~$ EXAMPLE_DIR=$(mktemp -d)

~$ cd $EXAMPLE_DIR

tmp.MEpbWK5rCB$ git init
Initialized empty Git repository in /tmp/tmp.MEpbWK5rCB/.git/

tmp.MEpbWK5rCB$ git submodule add git@github.com:awslabs/amazon-ecr-credential-helper.git
Cloning into '/tmp/tmp.MEpbWK5rCB/amazon-ecr-credential-helper'...
remote: Enumerating objects: 373, done.
remote: Counting objects: 100% (373/373), done.
remote: Compressing objects: 100% (300/300), done.
remote: Total 2186 (delta 90), reused 205 (delta 55), pack-reused 1813
Receiving objects: 100% (2186/2186), 2.60 MiB | 11.09 MiB/s, done.
Resolving deltas: 100% (889/889), done.

tmp.MEpbWK5rCB$ cd amazon-ecr-credential-helper/

amazon-ecr-credential-helper$ make docker
find: ‘.git/’: Not a directory
mkdir -p bin
docker run --rm \
-e TARGET_GOOS= \
-e TARGET_GOARCH= \
-v '/tmp/tmp.MEpbWK5rCB/amazon-ecr-credential-helper/bin':/go/src/github.com/awslabs/amazon-ecr-credential-helper/bin \
sha256:dfe7b32b42631a793e59eb8c18b99da8df91fa3717ecc4fa4d0323a1ccf84110
find: '.git/': Not a directory
git rev-parse --short=7 HEAD > GITCOMMIT_SHA
fatal: Not a git repository: ../.git/modules/amazon-ecr-credential-helper
make: *** [GITCOMMIT_SHA] Error 128
Makefile:75: recipe for target 'GITCOMMIT_SHA' failed
Makefile:32: recipe for target 'docker' failed
make: *** [docker] Error 2
```

[...]

PR #185, opened 2019-09-19
Fix `make docker` when repo is a submodule

scjudd:
*Issue #, if available:* #184 

*Description of changes:*

* Made `GITCOMMIT_SHA` a dependency of the `docker` Make target.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ DARWIN_AMD64_BINARY=bin/darwin-amd64/$(BINARY_NAME)
 WINDOWS_AMD64_BINARY=bin/windows-amd64/$(BINARY_NAME).exe
 
 .PHONY: docker
-docker: Dockerfile
+docker: Dockerfile GITCOMMIT_SHA
 	mkdir -p bin
 	docker run --rm \
 	-e TARGET_GOOS=$(TARGET_GOOS) \


Makefile
```diff
@@ -71,7 +70,7 @@ $(WINDOWS_AMD64_BINARY): $(SOURCES) GITCOMMIT_SHA
 	./scripts/build_variant.sh windows amd64 $(VERSION) $(shell cat GITCOMMIT_SHA)
 	@mv ./bin/windows-amd64/$(BINARY_NAME) ./$(WINDOWS_AMD64_BINARY)
 
-GITCOMMIT_SHA: $(GITFILES)
```
samuelkarp: Can you revert this change?  `$(GITFILES)` expresses a dependency in make so that the `GITCOMMIT_SHA` file gets regenerated any time there is a change to the files inside `.git`.  This ensures that `GITCOMMIT_SHA` gets updated when you make new commits, when you pull changes from a remote, and when you switch branches.

scjudd: Ah, that makes sense. Thanks for the explanation!

Could we not instead make this a `.PHONY` target and remove the `$(GITFILES) dependency? It would mean that it would run on _every_ `make`, but OTOH we're already checking everything in the `.git/` dir for changes on every `make` invocation, right?

samuelkarp: Making the `GITCOMMIT_SHA` target run on every `make` would mean that we'd end up rebuilding the binaries on each `make` invocation as well (since `GITCOMMIT_SHA` is a dependency of the binary targets).

PR #195, opened 2019-12-31
debian: updates for 0.3.1-1

samuelkarp:
*Description of changes:*

* Merged the v0.3.1 tag into the debian branch and removed the vendor folder.
* Updated the changelog
* Updated the control file to specify minimum version for the AWS SDK
* Removed outdated patches
* Updated the docker-deb script to support building from an upstream tarball

@nmeyerhans I have not yet uploaded a package to the mentors site, but can do so if this looks good.  As before, the Travis build will fail as dependencies are not present in the branch.


debian/changelog
```diff
@@ -1,4 +1,4 @@
-amazon-ecr-credential-helper (0.2.0-2) UNRELEASED; urgency=medium
+amazon-ecr-credential-helper (0.3.1-1) UNRELEASED; urgency=low
```
nmeyerhans: There should be an entry explicitly noting that this is a new upstream version. You don't need to go into any detail about what has actually changed, though if the new upstream version fixes any bugs that are open in Debian's BTS then this is where you'd indicate that. [Example](https://tracker.debian.org/media/packages/s/spamassassin/changelog-3.4.3-1)

debian/ORIG_URL
```diff
@@ -0,0 +1 @@
+https://amazon-ecr-credential-helper-releases.s3.us-east-2.amazonaws.com/0.3.1/release-novendor.tar.gz
```
nmeyerhans: Is there any reason not to grab the main release tarball and strip out the vendored components in the docker-deb script? Or, assuming the "upstream" releases are tagged in the repo, use `gbp export-orig` to construct the .orig.tar.xz file directly from the repo?

debian/changelog
samuelkarp: I need to include something like "New upstream release"?  What about the "urgency" level?

debian/ORIG_URL
samuelkarp: I think there are multiple valid ways to get a tarball.  The way I did this before was to use a make target in the upstream repo that generated a tarball that excluded the vendor folder and included a generated `GITCOMMIT_SHA` file and then used that.  This approach basically does the same thing, except that now the vendor-less tarball is actually published in a durable location and doesn't need to be regenerated.

I didn't really want to use the docker-deb script to modify the tarball itself as I didn't want the docker-deb script to be the only way to build this package.

As for `gbp export-orig`, I'm not very familiar with that.  Currently installing it in a container to go look at it more closely, but I don't know if it provides a mechanism to exclude the vendor folder or would give me a tarball that has the `GITCOMMIT_SHA` file.

debian/changelog
samuelkarp: Done

debian/ORIG_URL
nmeyerhans: You don't need `gbp export-orig` to export the final file; you can post-process it as needed.

debian/changelog
nmeyerhans: Urgency isn't really meaningful here.

debian/ORIG_URL
samuelkarp: Is there a preference for doing that over using an already-processed file that's published for this purpose?

nmeyerhans: It doesn't really matter. If the stripped version is useful beyond just the .deb package, then it's worth publishing upstream. If not, then it's pretty much a wash. The Debian archive will become the durable storage for the .orig file once it's uploaded, so so nothing is lost by not hosting it in s3.

samuelkarp: Got it.  The stripped version is already there and there's automation around producing it at the same time as the other release artifacts.  I also expect that other packagers who also don't want the vendor folder (for example, if this were to be packaged in Fedora) would find it useful.

At some point I plan to transition this project to using Go modules and away from vendoring entirely, so the difference between the full and the stripped tarball will go away.

samuelkarp:
@nmeyerhans Thank you!  Before I merge this: do I need to change `UNRELEASED` to `unstable` in the changelog?

PR #183, opened 2019-09-18
Bump AWS SDK to support sts:AssumeRoleWithWebIdentity

siwyd:
*Description of changes:*

In order to support native AWS IAM integration with Kubernetes,
the AWS SDK needs to be bumped to v1.23.13 at least. This PR
bumps the AWS SDK to v1.24.1.

More info: https://aws.amazon.com/blogs/opensource/introducing-fine-grained-iam-roles-service-accounts/

Action performed: dep ensure -update github.com/aws/aws-sdk-go

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #177, opened 2019-07-23
Update dependencies

bdwyertech:
*Description of changes:*

Updates dependencies, primarily to pull in the latest AWS-SDK-Go for various AssumeRole fixes.

https://github.com/aws/aws-sdk-go/commit/8be2a091080ce9f4d97f7018afca58cf0e28e3ff

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #197, opened 2020-01-02
Version 0.4.0

samuelkarp:
*Description of changes:*
* Feature - Added support for chaining assumed roles in the shared config file (`~/.aws/config`) defined by `source_profile` and `credential_source` ([#177](https://github.com/awslabs/amazon-ecr-credential-helper/pull/177))
* Feature - Added support for Web Identities and [IAM Roles for Service Accounts (IRSA)](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) with Kubernetes ([#183](https://github.com/awslabs/amazon-ecr-credential-helper/pull/183)) 
* Bug - Fixed the `make docker` target when the credential helper git repository is used as a git submodule ([#184](https://github.com/awslabs/amazon-ecr-credential-helper/issues/184))


docs/docker-credential-ecr-login.1
```diff
@@ -90,6 +104,11 @@ This command is not implemented and is a no-op.
 .TP 16
 .BR erase
 This command is not implemented and is a no-op.
+.SH BUGS
+Kubernetes users using the credential helper inside containers with a non-root
+user may encounter permission issues described here:
+https://github.com/kubernetes-sigs/external-dns/pull/1185.  You may be able to
+work around this bug by adjusting the Kubernetes \fIsecurityContext\fP.
 .SH COPYRIGHT
 Copyright 2018 Amazon.com, Inc. or its affiliates.  All rights reserved.
```
jahkeup: Should this be bumped now that we've hit 2020 or does this date not represent the "present" or last file-changed year? :fireworks: :calendar:

README.md
```diff
@@ -13,21 +13,8 @@ for the Docker daemon that makes it easier to use
...
-
-The Amazon ECR Docker Credential Helper uses the same credentials as the AWS
-CLI and the AWS SDKs. For more information about configuring AWS credentials,
-see
-[Configuration and Credential Files](http://docs.aws.amazon.com/cli/latest/userguide/cli-chap-getting-started.html#cli-config-files)
-in the *AWS Command Line Interface User Guide*.
-
-The credentials must have a policy applied that
-[allows access to Amazon ECR](http://docs.aws.amazon.com/AmazonECR/latest/userguide/ecr_managed_policies.html).
+You also must have AWS credentials available.  See the [AWS credentials section](#aws-credentials) for details on how to
```
jahkeup: Good call on breaking this out into its own section, I think it's more approachable written this way.

README.md
```diff
@@ -172,6 +161,40 @@ ECR registry:
 This is useful if you use `docker` to operate on registries that use different
 authentication credentials.
 
+### AWS credentials
+
+The Amazon ECR Docker Credential Helper allows you to use AWS credentials stored different locations.  Standard ones
```
jahkeup: nit: I think this is missing a word

```suggestion
The Amazon ECR Docker Credential Helper allows you to use AWS credentials stored in different locations.  Standard ones
```

docs/docker-credential-ecr-login.1
samuelkarp: Changed to "2018-2020".

README.md
samuelkarp: Fixed, thanks!

README.md
```diff
@@ -172,6 +161,40 @@ ECR registry:
...
+
+* Assumed roles specified with `role_arn` and `source_profile`
+* External credential processes specified with `credential_process`
+* Web Identities like [IAM Roles for Service Accounts in
+  Kubernetes](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) (*Note: Kubernetes
+  users using containers with a non-root user may encounter permission issues described in [this
+  bug](https://github.com/kubernetes-sigs/external-dns/pull/1185) and may need to employ a workaround adjusting the
+  Kubernetes `securityContext`.*)
+
+The Amazon ECR Docker Credential Helper uses the same credentials as the AWS
```
IRCody: This covers the credentials being used with the same precedence as the aws cli?

samuelkarp: Roughly yes, but the precise statement is that the credentials are used with the same precedence as the AWS SDK for Go.  The SDK attempts to match the CLI, but is sometimes imperfect at doing so (and some of the changes in this release are bugfixes for that).

IRCody: If we would consider it a bug if we don't match the CLI I think it's fine. If not, is it worth pointing out any known differences and/or pointing to the "Specifying Credentails" section here? https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/configuring-sdk.html.

samuelkarp: The CLI docs are more "user-facing" than the SDK docs since the CLI covers an application (the SDK docs cover how to write code using the SDK as well).  One of the major known differences is already pointed out in this section, specifically the part about `AWS_SDK_LOAD_CONFIG` being required when any value is to be read from the shared config file (`~/.aws/config`).

If you don't feel strongly about this, maybe we can move this to a separate PR/issue after this release goes out?

PR #201, opened 2020-02-04
api: enable shared config by default

### LINKED ISSUE #198: Enable shared config by default
Currently, the behavior of the AWS SDK for Go is to ignore the shared config file (`~/.aws/config`) unless explicitly enabled and the credential helper inherits this behavior.  The shared config file is not explicitly enabled in the credential helper today, but can be enabled by setting the environment variable `AWS_SDK_LOAD_CONFIG=true`.  It would be better to enable the shared config by default, and allow it to be disabled by `AWS_SDK_LOAD_CONFIG=false` if desired.

andrewsharon:
Fixes #198 

*Description of changes:*
The AWS SDK for Go ignores the shared config file `~/.aws/config` by
default, which is the inherited behavior of
amazon-ecr-credential-helper. This commit changes the default behavior
to have `~/.aws/config` enabled.

The shared config file usage can be controlled through the environment
variable `AWS_SDK_LOAD_CONFIG`. To disable this behavior, set
`AWS_SDK_LOAD_CONFIG=false`.

### DIFF
--- a/ecr-login/api/factory.go
+++ b/ecr-login/api/factory.go
@@ -14,6 +14,9 @@
 package api
 
 import (
+	"os"
+	"strconv"
+
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/endpoints"
 	"github.com/aws/aws-sdk-go/aws/request"
@@ -49,7 +52,10 @@ var userAgentHandler = request.NamedHandler{
 
 // NewClientWithDefaults creates the client and defaults region
 func (defaultClientFactory DefaultClientFactory) NewClientWithDefaults() Client {
-	awsSession := session.New()
+	awsSession := session.Must(session.NewSessionWithOptions(session.Options{
+		SharedConfigState: loadSharedConfigState(),
+	}))
+
 	awsSession.Handlers.Build.PushBackNamed(userAgentHandler)
 	awsConfig := awsSession.Config
 	return defaultClientFactory.NewClientWithOptions(Options{
@@ -60,7 +66,10 @@ func (defaultClientFactory DefaultClientFactory) NewClientWithDefaults() Client 
 
 // NewClientWithFipsEndpoint overrides the default ECR service endpoint in a given region to use the FIPS endpoint
 func (defaultClientFactory DefaultClientFactory) NewClientWithFipsEndpoint(region string) (Client, error) {
-	awsSession := session.New()
+	awsSession := session.Must(session.NewSessionWithOptions(session.Options{
+		SharedConfigState: loadSharedConfigState(),
+	}))
+
 	awsSession.Handlers.Build.PushBackNamed(userAgentHandler)
 
 	endpoint, err := getServiceEndpoint("ecr-fips", region)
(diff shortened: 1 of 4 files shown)


ecr-login/api/factory.go
```diff
@@ -109,3 +120,12 @@ func getServiceEndpoint(service, region string) (string, error) {
...
 	return endpoint.URL, err
 }
+
+func loadSharedConfigState() session.SharedConfigState {
+	awsSdkLoadConfig := os.Getenv("AWS_SDK_LOAD_CONFIG")
+	enableSharedConfig, _ := strconv.ParseBool(awsSdkLoadConfig)
+	if !enableSharedConfig && awsSdkLoadConfig != "" {
+		return session.SharedConfigDisable
+	}
+	return session.SharedConfigEnable
```
samuelkarp: I'd recommend a change like this:

```go
loadConfig := os.Getenv("AWS_SDK_LOAD_CONFIG")
if loadConfig == "" {
	return session.SharedConfigEnable
}
if enable, err := strconv.ParseBool(loadConfig); err == nil && !enable {
	return session.SharedConfigDisable
}
return session.SharedConfig
```

This gives you:

* Slightly shorter variable names (generally good in Go, especially in short functions)
* Removing ambiguity around "awsSdk" vs "awsSDK" (should be [capitalized in Go](https://github.com/golang/go/wiki/CodeReviewComments#initialisms))
* Errors not ignored
* More explicit that nothing defined means enabled and only explicitly setting to false is disabled

ecr-login/api/factory_test.go
```diff
@@ -0,0 +1,45 @@
...
+import (
+	"os"
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws/session"
+	"github.com/stretchr/testify/assert"
+)
+
+func TestSharedConfigStateDefaultBehavior(t *testing.T) {
+	os.Unsetenv("AWS_SDK_LOAD_CONFIG") // Ensure the environment variable is not set
```
samuelkarp: For each of these, I would recommend reading and restoring the original environment variable state.

You could also convert this into a single table-driven test (with Go subtests) since they have the same setup logic and assertions.

```go
const loadConfigEnvVar = "AWS_SDK_LOAD_CONFIG"

func TestSharedConfigState(t *testing.T) {
	defer func(value string) {
		os.Setenv(loadConfigEnvVar)
	}(os.Getenv(loadConfigEnvVar)

	cases := []struct{
		envValue string
		expected session.SharedConfigState
	}{
		{"", session.SharedConfigEnable},
		{"true", session.SharedConfigEnable},
		{"false", session.SharedConfigEnable},
	}

	for _, testCase := range cases {
		t.Run(testCase.envValue, func(t *testing.T) {
			os.Setenv(loadConfigEnvVar, testCase.envValue)
			state := loadSharedConfigState()
			assert.NotNil(t, state)
			assert.Equal(t, testCase.expected, state)
		})
	}
}

andrewsharon:
Thanks for the feedback! I've made the changes and updated the PR.

samuelkarp:
This code looks good to me; just waiting for the Travis build to complete before approving.

PR #208, opened 2020-04-19
Use the XDG Basedir Spec for the cache

### LINKED ISSUE #205: Respect XDG Base Directory specification for cache files
HI,

I wish that the `.ecr` directory would be located in `$XDG_CACHE_HOME/ecr` if the env var is set. This is a good reference: https://wiki.archlinux.org/index.php/XDG_Base_Directory

Thanks

WhyNotHugo:
Currently, for caching data, a new directory is created in the current user's home directory. This is bad practice (if all apps just dump files into the user's home, it becomes somewhat unmaintainable).

The XDG Basedir Specification covers this pretty neatly; the cache directory for all applications is configurable via the `XDG_CACHE_HOME` variable, and falls back to `~/.cache` if unset.

This PR makes the cache for all applications live in one single location, rather that littering the user's home.

Fixes #205

### DIFF
--- a/README.md
+++ b/README.md
@@ -214,7 +214,8 @@ There is no need to use `docker login` or `docker logout`.
 
 ## Troubleshooting
 
-Logs from the Amazon ECR Docker Credential Helper are stored in `~/.ecr/log`.
+Logs from the Amazon ECR Docker Credential Helper are stored inside you cache
+files directory (e.g.: `~/.cache/ecr/log` on *nix).
 
 For more information about Amazon ECR, see the the
 [Amazon Elastic Container Registry User Guide](http://docs.aws.amazon.com/AmazonECR/latest/userguide/what-is-ecr.html).
--- a/ecr-login/config/cache_dir.go
+++ b/ecr-login/config/cache_dir.go
@@ -19,9 +19,7 @@ func GetCacheDir() string {
 	if cacheDir := os.Getenv("AWS_ECR_CACHE_DIR"); cacheDir != "" {
 		return cacheDir
 	}
-	xdgDir := os.Getenv("XDG_CACHE_HOME");
-	if xdgDir == "" {
-		xdgDir = "~/.cache"
-	}
-	return xdgDir + "/ecr"
+	dir, _ := os.UserCacheDir()
+
+	return dir + "/ecr"
 }
--- a/ecr-login/config/cache_dir.go
+++ b/ecr-login/config/cache_dir.go
@@ -19,5 +19,9 @@ func GetCacheDir() string {
 	if cacheDir := os.Getenv("AWS_ECR_CACHE_DIR"); cacheDir != "" {
 		return cacheDir
 	}
-	return "~/.ecr"
+	xdgDir := os.Getenv("XDG_CACHE_HOME");
+	if xdgDir == "" {
+		xdgDir = "~/.cache"
+	}
+	return xdgDir + "/ecr"
 }


samuelkarp:
Hello @WhyNotHugo, thanks for sending this pull request!  There are a few things that I think need to be addressed before we can merge a change like this:

* This change as it stands is not backwards-compatible with installations where files are already present in `~/.ecr`.  The credential helper should either continue to use those files when present, or handle migrating them to the new location without intervention.
* The [XDG Base Directory Specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) is widely used on Linux, but not so on other operating systems.  The credential helper supports Linux, Windows, and macOS, and should work with the appropriate location for each.  See https://github.com/golang/go/issues/22536 for a discussion of the correct cache location, and https://github.com/golang/go/issues/29960 for a discussion of config locations.  The code as-is doesn't use the correct locations for Windows or macOS, but I would prefer to not migrate twice; we should either gate this change to be Linux-only or also move to the correct locations for Windows and macOS at the same time.  Go 1.11 added [`os.UserCacheDir`](https://golang.org/pkg/os/#UserCacheDir); I would be okay dropping support for Go 1.10 in order to use that function.
* This change affects the log file location, which is documented in the README; that needs to be updated.
* Since the file is moving, I think the new location should be documented in the README and the [manual page](https://github.com/awslabs/amazon-ecr-credential-helper/blob/master/docs/docker-credential-ecr-login.1).

If you're willing and able to make these changes, I'd be happy to move forward with this PR.  If you're not, I'd be happy to leave this open and then use it as the base of a change that does address the other concerns.

WhyNotHugo:
Hi! I was unaware that Mac used a different location (should have guessed it!). Thanks for pointing out that to already has a function that does the heavy lifting -- I'm very happy that an external lib is not needed.

I'll address these changes and update the PR. It'll take me a while since I'm very rusty in go, but it's a fair challenge! :)

Thanks!

WhyNotHugo:
[Note: Still have pending moving existing files]

> Since the file is moving, I think the new location should be documented in the README and the manual page.

There seems to be no mention on the manual page right now. Should I add a FILES section at the bottom?

samuelkarp:
> There seems to be no mention on the manual page right now. Should I add a FILES section at the bottom?

You're right!  It would be nice to add a FILES section, but I won't block accepting this PR on that change.

ecr-login/config/cache_dir.go
```diff
@@ -19,5 +19,7 @@ func GetCacheDir() string {
 	if cacheDir := os.Getenv("AWS_ECR_CACHE_DIR"); cacheDir != "" {
 		return cacheDir
 	}
-	return "~/.ecr"
+	dir, _ := os.UserCacheDir()
```
samuelkarp, line 22: Let's not ignore the error here.

```suggestion
	dir, err := os.UserCacheDir()
	if err != nil {
		return "~/.ecr"
	}
```

README.md
```diff
@@ -214,7 +214,8 @@ There is no need to use `docker login` or `docker logout`.
 
 ## Troubleshooting
 
-Logs from the Amazon ECR Docker Credential Helper are stored in `~/.ecr/log`.
+Logs from the Amazon ECR Docker Credential Helper are stored inside you cache
+files directory (e.g.: `~/.cache/ecr/log` on *nix).
```
samuelkarp, line 218: Let's either link to the [Go documentation](https://golang.org/pkg/os/#UserCacheDir) or describe the specific locations for Linux, Windows, and Mac.

samuelkarp:
Hey @WhyNotHugo! I've taken your commits and adjusted them to add fall-back behavior to `~/.ecr` when it's present so that people upgrading can continue to use the existing cache.  The new PR can be found here: https://github.com/awslabs/amazon-ecr-credential-helper/pull/250.

WhyNotHugo:
Sorry, I never got around to finishing those changes (plus, my `go` is super rusty).

Happy that it's being picked up. I'm of course perfectly fine with the code being reused! Cheers!

PR #221, opened 2020-06-22
Switch from dep to Go modules

kzys:
*Issue #, if available:*

NA

*Description of changes:*

Go 1.15 is coming. Even Amazon Linux 2 is having Go 1.13. We should be
able to use Go modules now.


kzys:
Let me move `vendor/` to `ecs-login/` as like before.

ecr-login/vendor/golang.org/x/sys/unix/mkasm_darwin.go
```diff
@@ -1,61 +0,0 @@
-// Copyright 2018 The Go Authors. All rights reserved.
-// Use of this source code is governed by a BSD-style
-// license that can be found in the LICENSE file.
-
-// +build ignore
-
-// mkasm_darwin.go generates assembly trampolines to call libSystem routines from Go.
-//This program must be run after mksyscall.go.
```
kzys, line 8: These removed files have "+build ignore" and apparently used as code generators. Removing them shouldn't affect this package.

kzys: https://github.com/golang/sys/blob/master/unix/README.md#mksyscallgo

Issue #181, opened 2019-09-13
Support AssumeRole when `credential_source = Environment`


b-willard:
**Use Case**

I am storing credentials (`AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`) encrypted in a remote vault and do not want to expose them to the docker host in question via ~/.aws/credentials. Instead, credentials should temporarily exist on the docker host via the environment. The environment is populated during a configuration management run or image build provisioning process which interfaces with the remote vault and decrypts the credentials.

**Issue**

The configuration required to meet this use case results in a catch-22:

~/.aws/config with [environment credential source][2]
```ini
[profile some_ecr_readonly_profile]
region = us-west-2
role_arn = arn:aws:iam::xxxxxxxxxxxx:role/Some-ECR-ReadOnly-Role
credential_source = Environment
```
environment vars
```
AWS_SDK_LOAD_CONFIG = '1'
AWS_PROFILE = 'some_ecr_readonly_profile'
AWS_ACCESS_KEY_ID = 'XXXXXXXXXXXXXXXXXXXX'
AWS_SECRET_ACCESS_KEY = 'XXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX'
```
where `arn:aws:iam::xxxxxxxxxxxx:role/Some-ECR-ReadOnly-Role` uses the [AWS ECR Read Only managed policy][3] and the IAM user associated to `AWS_ACCESS_KEY_ID` has the following policy directly attached:
```json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Sid": "1",
            "Effect": "Allow",
            "Action": "sts:AssumeRole",
            "Resource": "arn:aws:iam::xxxxxxxxxxxx:role/Some-ECR-ReadOnly-Role"
        }
    ]
}
```

With this setup and aws-go-sdk's credential load precedence, AWS_ACCESS_KEY_ID and AWS_SECRET_ACCESS_KEY are used directly rather than AssumeRole'ing and using the resulting STS credentials instead. As a result, any `docker pull` or `docker push` under this configuration results in #164. This is acknowledged with a solution in [the sdk's documentation][1].

**Solution**

I have a working `docker-credential-ecr-login` binary meeting my needs based on the diff below. (I did not submit this as an official PR because these are the first lines of Go I've ever written :D so I'm certain there are better methods.) The change essentially asks two questions:

[...]

PR #231, opened 2020-08-04
Support AssumeRole when `credential_source = Environment`

benhowes:
From #181 

*Description of changes:*
This PR is a slight variation on the code outlined in #181, which just always passes the value of `AWS_PROFILE` to the creation of a session. To the best of my knowledge, this will work in the regular cases which are already supported, as well as when assuming a role with `credential_source = Environment`.

I've not currently added any tests, because I cannot see any examples of tests which use the `~/.aws/config` or any tests which set `AWS_PROFILE`. 

Opening this for discussion and to hopefully arrive at a solution which is mergeable.

### DIFF
--- a/ecr-login/ecr.go
+++ b/ecr-login/ecr.go
@@ -16,6 +16,7 @@ package ecr
 import (
 	"errors"
 	"fmt"
+	"os"
 
 	"github.com/sirupsen/logrus"
 
@@ -52,6 +53,8 @@ func (self ECRHelper) Get(serverURL string) (string, string, error) {
 		return "", "", credentials.NewErrCredentialsNotFound()
 	}
 
+	profile, profile_exists := os.LookupEnv("AWS_PROFILE")
+
 	var client api.Client
 	if registry.FIPS {
 		client, err = self.ClientFactory.NewClientWithFipsEndpoint(registry.Region)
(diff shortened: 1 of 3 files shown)


ecr-login/ecr.go
```diff
@@ -52,13 +53,21 @@ func (self ECRHelper) Get(serverURL string) (string, string, error) {
 		return "", "", credentials.NewErrCredentialsNotFound()
 	}
 
+	profile, profile_exists := os.LookupEnv("AWS_PROFILE")
```
kzys, line 56: nit: profile_exists -> profileExists

https://golang.org/doc/effective_go.html#mixed-caps

>  Finally, the convention in Go is to use MixedCaps or mixedCaps rather than underscores to write multiword names.

ecr-login/ecr.go
```diff
@@ -52,13 +53,21 @@ func (self ECRHelper) Get(serverURL string) (string, string, error) {
...
 		if err != nil {
 			logrus.WithError(err).Error("Error resolving FIPS endpoint")
 			return "", "", credentials.NewErrCredentialsNotFound()
 		}
+	} else if profile_exists {
+		client, err = self.ClientFactory.NewClientWithExplicitProfile(profile)
+		if err != nil {
+            logrus.WithError(err).Error("Error creating client with explicit profile")
+            return "", "", credentials.NewErrCredentialsNotFound()
+        }
```
kzys, line 70: This is probably a question for @samuelkarp - Right now FIPS and AWS_PROFILE are mutually-exclusive. Is there a case where a customer want to use AWS_PROFILE, but with FIPS?

kzys (changes requested):

ecr-login/ecr.go
samuelkarp: Yes, these options should not be mutually exclusive.

samuelkarp:
@benhowes Are you able to update this to address the mutual-exclusivity problem?

benhowes:
@samuelkarp I've actually left the company where I was developing this since then, I don't think I'll find time to work on this PR any further. Sorry!

samuelkarp:
@benhowes Thanks for letting me know!  I'm going to close this PR for now, but we'd be open to looking at changes like this again in the future.

PR #237, opened 2020-09-01
README: Obvious string replacement for ECR URI

ryansch:
*Description of changes:*
I ran into an issue with my coworkers today while onboarding our ECR environment where it wasn't obvious that the region needed to be set in their docker config.

We ended up with the failure mode that they had a literal `region` in their config instead of something like `us-east-1`.

I've wrapped the two things that need to be replaced with `<>`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -156,7 +156,7 @@ ECR registry:
 ```json
 {
 	"credHelpers": {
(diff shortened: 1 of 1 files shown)


kzys (approved):
Looks good to me. @samuelkarp What do you think?

samuelkarp (approved):

PR #244, opened 2020-11-12
Use GitHub Actions

kzys:
Signed-off-by: Kazuyoshi Kato <katokazu@amazon.com>

*Issue #, if available:*

*Description of changes:*


kzys:
I'm unsure this PR which enables GitHub Actions can get the results from GitHub Actions. Seems like it was though.

https://github.com/firecracker-microvm/firecracker-containerd/pull/404/checks

.github/workflows/build.yaml
```diff
@@ -0,0 +1,32 @@
...
+    env:
+      GO111MODULE: on
+
+    # Intentionally use 18.04 instead of "latest" to
+    # make this build reproducible.
+    runs-on: ubuntu-18.04
+
+    strategy:
+      matrix:
+        go: ['1.10', '1.11', '1.12', '1.13', '1.14']
```
samuelkarp, line 20: Let's drop 1.10 and add 1.15.

kzys: Sure. Did that in the last revision.

kzys:
I'm unsure about the value of Travis CI macOS builds. We cross-compile the helper for Windows and macOS even in Linux. Would the real macOS builds unveil something we cannot reproduce in the cross-compiled builds?

samuelkarp:
> Would the real macOS builds unveil something we cannot reproduce in the cross-compiled builds?

We can't run unit tests in cross-compiled builds, so it'd be good to have Mac and Windows unit tests running.  GitHub Actions has runners for both Mac and Windows too, so I think we can get rid of Travis completely.

kzys:
Wow I didn't know it has macOS. Let me include that as well.

Makefile
```diff
@@ -45,7 +45,7 @@ $(LOCAL_BINARY): $(SOURCES) GITCOMMIT_SHA
 
 .PHONY: test
 test:
-	. ./scripts/shared_env && go test -v -timeout 30s -short -cover $(shell go list ./ecr-login/... | grep -v /vendor/)
```
samuelkarp, line 48: Are vendored dependencies excluded now?

kzys: Yes. Seems Go can recognize the vendored dependencies as their original names, not `github.com/awslabs/amazon-ecr-credential-helper/ecr-login/vendor/github.com/...`. So it doesn't match `./...`.

* Go 1.11: https://github.com/awslabs/amazon-ecr-credential-helper/pull/244/checks?check_run_id=1393304202
* Go 1.15: https://github.com/awslabs/amazon-ecr-credential-helper/pull/244/checks?check_run_id=1393304305

samuelkarp (approved):

PR #250, opened 2020-12-22
[Carry #208]  Use the XDG Basedir Spec for the cache

### LINKED ISSUE #205: Respect XDG Base Directory specification for cache files

samuelkarp:
*Issue #, if available:*
Fixes #205
Carries #208 (from @WhyNotHugo)

*Description of changes:*
Currently, for caching data, a new directory is created in the current user's home directory. This is bad practice (if all apps just dump files into the user's home, it becomes somewhat unmaintainable).

The XDG Basedir Specification covers this pretty neatly; the cache directory for all applications is configurable via the `XDG_CACHE_HOME` variable, and falls back to `~/.cache` if unset.

The os.UserCacheDir function returns the per-platorm cache dir, since the XDG-Basedir is *nix-specific (except macOS).

This changeset makes the cache for all applications live in one single location, rather that littering the user's home.

*Testing:*

Manual testing covers the behavior when `~/.ecr` exists and when it does not.

* [x] Linux
* [ ] Windows
* [ ] MacOS


ecr-login/cache/build.go
```diff
@@ -34,13 +33,12 @@ func BuildCredentialsCache(awsSession *session.Session, region string, cacheDir
 
 	if cacheDir == "" {
 		//Get cacheDir from env var "AWS_ECR_CACHE_DIR" or set to default
```
kzys, line 35: GetCacheDir does more than what this comment explains. How about removing the comment from here?

Issue #248, opened 2020-12-07
[Feature Request] Add Support for Public ECR


PatrickXYS:
Given we have launched [Public ECR registry](https://aws.amazon.com/blogs/aws/new-amazon-ec2-g4ad-instances-featuring-amd-gpus-for-graphics-workloads/?nc1=b_nrp) last week, we should expect huge volume of feature-requests for using amazon-ecr-credential-helper with public ecr support added.

We're experimenting POC for CD pipeline by using amazon-ecr-credential-helper, and expect to massively use public ECR.

I can bring up a design doc and implement Public ECR alongside with existing Private ECR implementation.

PatrickXYS:
/cc @samuelkarp @kzys 

Appreciate any insights from you!

Jeffwan:
Seems ecr internally doesn't handle public registry. Instead, a separate API is used. 

https://docs.aws.amazon.com/sdk-for-go/api/service/ecrpublic/

To give the best user experience, I think it's great to support public ecr in this credential helper as well rather than creating separate one. It should determine the api to use based on the registry patterns internally and hide as many details as it can. 

- public: `public.ecr.aws/h4kxxxxx`
- private: `692xxxxxxx.dkr.ecr.us-east-1.amazonaws.com`

jahkeup:
Thanks for bringing this up! 

> Seems ecr internally doesn't handle public registry. Instead, a separate API is used.

You're exactly right, this means the credential helper will need to be refactored a bit to handle both ECR refs and ECR Public refs in order to provide the correct set of credentials to Docker. Today, the credential helper assumes that it either is or isn't responsible for ECR refs only and has a single factory that is configured based on the ref.

johnnyshields:
AWS team, this is a huge issue. I'm flabbergasted that ECR Public was launched without this. We switched all our builds from DockerHub to ECR Public only to be in a worse situation re: rate limiting. Builds are broken everywhere 🔥🔥🔥 

Please make sure cross-region support is included in this; we are hosting in `ap-northeast-1`

vivekpd15:
+1

PR #253, opened 2020-12-24
Support for ECR Public

samuelkarp:
*Issue #, if available:*
Fixes https://github.com/awslabs/amazon-ecr-credential-helper/issues/248
Depends on https://github.com/awslabs/amazon-ecr-credential-helper/pull/252

*Description of changes:*
This set of commits adds support for ECR Public.  Note that it depends on https://github.com/awslabs/amazon-ecr-credential-helper/pull/252, so you can skip those commits when reviewing this one (or they'll disappear from the diff when we merge that PR).

ECR Public credentials are cached and returned by the list command by default.

/cc @micahhausler


rpnguyen (approved):
Awesome, thanks for the quick turnaround, lgtm :shipit:

PatrickXYS:
@samuelkarp Any updates here, looks like the test failed for dependencies?

samuelkarp:
> Any updates here, looks like the test failed for dependencies?

@PatrickXYS Considering I've been on vacation since right after opening this PR, no.

I don't currently have access to a Mac.  It looks like an update to `golang.org/x/sys` is incompatible with Go 1.11 on Mac OS 10.15.  We'll need to figure out whether to roll-back the update to that dependency or drop support for Go 1.11 on Mac OS 10.15.

ecr-login/api/client.go
```diff
@@ -120,8 +152,42 @@ func (c *defaultClient) GetCredentialsByRegistryID(registryID string) (*Auth, er
...
+	if err != nil && cachedEntry != nil {
+		logrus.WithError(err).Info("Got error fetching authorization token. Falling back to cached token.")
+		return extractToken(cachedEntry.AuthorizationToken, cachedEntry.ProxyEndpoint)
+	}
+	return auth, err
+}
+
 func (c *defaultClient) ListCredentials() ([]*Auth, error) {
-	auths := []*Auth{}
+	// prime the cache
```
jahkeup, line 180:
```suggestion
	// prime the cache with default authorization tokens.
```

ecr-login/api/client.go
```diff
@@ -196,15 +251,41 @@ func (c *defaultClient) getAuthorizationToken(registryID string) (*Auth, error)
 	return nil, fmt.Errorf("No AuthorizationToken found for %s", registryID)
 }
 
+func (c *defaultClient) getPublicAuthorizationToken() (*Auth, error) {
+	var input *ecrpublic.GetAuthorizationTokenInput
+
+	output, err := c.ecrPublicClient.GetAuthorizationToken(input)
+	if err != nil {
+		return nil, errors.Wrap(err, "ecr: Failed to get authorization token")
```
jahkeup, line 259: nit: case
```suggestion
		return nil, errors.Wrap(err, "ecr: failed to get authorization token")
```

kzys:
I'm inclined to drop Go 1.11 support in macOS and other OSes.

* The Go team no longer supports Go 1.11.
* Debian 10 (buster) has Go 1.11  but newer versions (1.13 and 1.14) are available in buster-backports.
* Amazon Linux 2 has Go 1.13.

What do you think? Are there other places where customers have to use Go 1.11?

jahkeup (commented):
Still reading through, flushing minor nits for now.

samuelkarp:
> I'm inclined to drop Go 1.11 support in macOS and other OSes.

@kzys I'm comfortable with that proposal.  I _think_ the only thing that needs to be updated is the GitHub actions workflow as the readme already recommends 1.12+ (though possibly we could bump that up to 1.14+ as well).  My preference would be to have that change come in as a separate PR and then rebase this one on top.

ecr-login/api/client.go
```diff
@@ -20,43 +20,61 @@ import (
...
+	FIPS    bool
+	Region  string
 }
 
 // ExtractRegistry returns the ECR registry behind a given service endpoint
 func ExtractRegistry(serverURL string) (*Registry, error) {
 	if strings.HasPrefix(serverURL, proxyEndpointScheme) {
 		serverURL = strings.TrimPrefix(serverURL, proxyEndpointScheme)
 	}
+	if strings.HasPrefix(serverURL, ecrPublicName) {
```
samuelkarp, line 61: This check isn't sufficient as something like `public.ecr.aws.justkidding.example.com` would match.

kzys: ecrPattern is having the same problem as well. I think it would be better to parse `serverURL` as a URL first and check specific components (scheme, host, ..) instead.

samuelkarp: Fixed, @kzys can you take a look?

ecr-login/api/client.go
samuelkarp: Done

ecr-login/api/client.go
samuelkarp: Done

ecr-login/api/client.go
```diff
@@ -16,47 +16,70 @@ package api
...
+const (
+	proxyEndpointScheme = "https://"
+	programName         = "docker-credential-ecr-login"
+	ecrPublicName       = "public.ecr.aws"
+	ecrPublicEndpoint   = proxyEndpointScheme + ecrPublicName
 )
 
-const proxyEndpointScheme = "https://"
-const programName = "docker-credential-ecr-login"
+var ecrPattern = regexp.MustCompile(`(^[a-zA-Z0-9][a-zA-Z0-9-_]*)\.dkr\.ecr(\-fips)?\.([a-zA-Z0-9][a-zA-Z0-9-_]*)\.amazonaws\.com(\.cn)?$`)
```
kzys, line 40: `\-fips` -> `-fips`? `-` is not a meta character.

ecr-login/api/client.go
```diff
@@ -16,47 +16,70 @@ package api
...
+			Service: ServiceECRPublic,
+		}, nil
 	}
-	matches := ecrPattern.FindStringSubmatch(serverURL)
+	matches := ecrPattern.FindStringSubmatch(serverURL.Hostname())
 	if len(matches) == 0 {
 		return nil, fmt.Errorf(programName + " can only be used with Amazon Elastic Container Registry.")
 	} else if len(matches) < 3 {
-		return nil, fmt.Errorf(serverURL + "is not a valid repository URI for Amazon Elastic Container Registry.")
+		return nil, fmt.Errorf(input + "is not a valid repository URI for Amazon Elastic Container Registry.")
```
kzys, line 75: It might be better to have ` ` before "is not a valid ...". The current implementation results "foobaris not a valid ..."

Alternatively you can do "%q is not a valid ..."

ecr-login/api/client.go
samuelkarp: Done

ecr-login/api/client.go
samuelkarp: Done

kzys (approved):
LGTM!

PR #264, opened 2021-02-15
debian: update packaging for 0.5.0-1

samuelkarp:
*Description of changes:*
debian: update packaging for 0.5.0-1


debian/control
```diff
@@ -7,12 +7,11 @@ Build-Depends:
  dh-golang (>= 1.1),
  golang-go (>= 1.7),
  git,
- golang-github-aws-aws-sdk-go-dev (>= 1.16.26),
+ golang-github-aws-aws-sdk-go-dev (>= 1.36.0),
```
jahkeup, line 10: I wonder if there's tooling to align the required packages with the `go.mod` specified versions (which I'm guessing is why this was force-pushed to [match 1.36.0](https://github.com/awslabs/amazon-ecr-credential-helper/pull/264/commits/390d97528616b4b5a507b1d871c1d49380fb4edc) instead of 1.35.4). This would have been an easy oversight and could have slipped through.

Maybe @nmeyerhans would know if such tooling exists?

jahkeup (approved):

debian/control
samuelkarp: > which I'm guessing is why this was force-pushed to match 1.36.0 instead of 1.35.4

:sweat_smile: yes...I had most of this top commit written prior to determining whether the release would include https://github.com/awslabs/amazon-ecr-credential-helper/pull/253, which is why I had the mismatch...

samuelkarp: Note that unlike with Go modules, the dependencies here follow Debian's/Apt's style of version matching.  That means that even though the dependency was specified as `>= 1.35.4`, it was actually being built against [1.36.33-1](https://packages.debian.org/source/unstable/golang-github-aws-aws-sdk-go) (the latest present in `unstable`, since that's where I do my test builds).

nmeyerhans (approved):

kzys (approved):

nmeyerhans:
I just noticed a couple issues:

1. Is the Go build-dependency right? Will the cred helper actually build with Go 1.7? If so, it's fine, but if not it could cause issues for people who want to backport the package if the version restrictions aren't actually accurate.
2. After performing a build, then running fakeroot ./debian/rules clean, I still have a bin/local/docker-credential-ecr-login file. It should get cleaned up by the clean rule.

samuelkarp:
> Is the Go build-dependency right? Will the cred helper actually build with Go 1.7? If so, it's fine, but if not it could cause issues for people who want to backport the package if the version restrictions aren't actually accurate.

No, the minimum is now Go 1.12.

I'll have an update addressing both of these issues shortly.

jahkeup (approved):

nmeyerhans (approved):

PR #285, opened 2021-07-19
Migrate to AWS SDK v2

### LINKED ISSUE #282: Update to AWS SDK v2
Could we update this tool to use v2 of aws-sdk-go? Most of all I would like to be able to assume roles via EC2 instance metadata (this works on v2 but not v1).

christopher-dG:
*Issue #, if available:* 

Closes #282, I think it might also solve a few other auth-related issues in this repo.

*Description of changes:*

- Migrated to AWS SDK v2.
- The check for `AWS_SDK_LOAD_CONFIG` is gone, since I couldn't find any mention it in SDK v2. It's always enabled now.
- I'm not sure if I correctly "translated" the endpoint resolving/discovery, since I'm really unfamiliar with that as a whole.

I have confirmed this works for my own use cases (assuming a role via EC2 instance metadata with a config profile).


kzys (approved):
Looks good to me.

@samuelkarp What do you think?

christopher-dG:
Oh by the way, I think the third party license file needs to be updated, I wasn't entirely sure how to format it.

omus:
Bump

kzys:
I can take a look the license file part. Can you squash the vendor commit and the revendor commit into one? They don't have to be separate, I think.

I'm going to remove Go 1.12. The version is no longer supported by the Go team.

Issue #290, opened 2021-08-10
New release that supports AWS SSO


oofnikj:
Since #229 was resolved several months ago, I'd like to request cutting a new release of this tool with the improvements merged since. It's not always convenient or possible to build from source.

sergeylanzman:
@kzys @samuelkarp 
Please release new version of  `amazon-ecr-credential-helper`. 
Last release was on Feb 15... 
SSO support it's huge blocker for switch use AWS sso in our company.

tomaszdudek7:
We need that desperately.

rlangfordBV:
New release please :)

whyman:
Pretty please? With a 🍒 on top?

dancmeyers:
This would be amazing!

jessebye:
The PR was merged in March. That makes it about 6 months now with no releases. Has this project been abandoned? Is it time to fork and publish our own version?

kyledecot:
🤞 for a new release.

jessebye:
If it helps anyone, I've forked this repo and created a new release that should support SSO:
https://github.com/jessebye/amazon-ecr-credential-helper/releases/tag/v0.6.0

calebwhitt:
Thank you @jessebye!

Would be great to know if the original amazon-ecr-credential-helper has been abandoned too. Any updates?

kzys:
Sorry for the very late reply. No, the helper hasn't been abandoned. We are going to have a new release to include unreleased changes.

That being said, we are internally discussing the model to support the helper more proactively. We've been slow to respond PRs and issues and we'd like to change that.

calebwhitt:
Thanks for the update @kzys, much appreciated :)

daniel-farrell-blockfi:
I was able to rebuild this and get it working.  If anyone wants the binary I ended up with for `darwin`, it's at https://github.com/daniel-farrell-blockfi/amazon-ecr-credential-helper/releases/tag/0.6.3

icorbett:
Is there any reason not to tag a new release during the time that internal discussions are happening? This project seems to be quite stable even if its future is less certain. With SSO being increasingly common, this seems like an unnecessary stumbling block.

jmriebold:
This is an issue for us as well, as our whole organization has just switched to SSO.

jobimrobinsantos-drizly:
[...]

PR #307, opened 2022-01-18
spec: update for 0.6.0-1

austinvazquez:
*Issue #, if available:*
#290 

*Description of changes:*
Update spec for release 0.6.0-1


amazon-ecr-credential-helper.spec
```diff
@@ -118,6 +136,11 @@ install -D -m 0644 \
 rm -rf %{buildroot}
 
 %changelog
+* Fri Jan 14 2022 Austin Vazquez <macedonv@amazon.com> - 0.6.0-1
+- Added support for AWS SSO
+- Added support to assume roles via EC2 instance metadata
+- Added support for Apple Silicon
```
kzys, line 142: @samuelkarp Did you "curate" ChangeLog entries before or was is just as is? I wonder whether we should mention Apple Silicon or not.

austinvazquez: FWIW, this wording is inline with the release notes [here](https://github.com/awslabs/amazon-ecr-credential-helper/blob/main/CHANGELOG.md#060).

kzys:
Okay. This shouldn't be a blocker. The Amazon Linux folks could ping us if the spec file doesn't meet their bar.

amazon-ecr-credential-helper.spec
samuelkarp: > Did you "curate" ChangeLog entries before

Yes, generally the entries in the RPM changelog only cover what's relevant to the RPM.

austinvazquez: Should be resolved in latest. 👍🏼

PR #309, opened 2022-02-09
Allow callers to set log output

imjasonh:
Signed-off-by: Jason Hall <jasonhall@redhat.com>

*Description of changes:*

Though this code is typically consumed as a CLI cred helper configured for use with the `docker` CLI, we've also found it to be really useful as a Go client for generating ECR credentials from the environment.

[`go-containerregistry`](https://github.com/google/go-containerregistry), which is used by [Tekton](https://tekton.dev), [Knative](https://knative.dev), [`ko`](https://github.com/google/ko), [`cosign`](https://github.com/sigstore/cosign), [Kaniko](https://kaniko.dev), [Buildpacks](https://buildpacks.io) and more, provides an adapter between its auth-providing [`authn.Keychain`](https://godoc.org/github.com/google/go-containerregistry/pkg/authn#Keychain) interface, and Docker's `credential.Helper` interface, which you implement here. This means that these tools can (and some already do!) embed the cred helper's logic with:

```go
kc := authn.NewKeychainFromHelper(ecr.ECRHelper{ClientFactory: api.DefaultClientFactory{}})
```

For example:
- https://github.com/google/ko/blob/89ede9110a1333a31bf417cb5fed6f69f8084ff1/pkg/commands/config.go#L45
- https://github.com/GoogleContainerTools/kaniko/blob/a7425d1fd0442b58dc24698285102176365a28d9/pkg/creds/creds.go#L32
- https://github.com/sigstore/cosign/blob/22007e56aee419ae361c9f021869a30e9ae7be03/cmd/cosign/cli/options/registry.go#L78

And this works great! 🎉  These clients don't have to install the cred helper CLI and configure it using `~/.docker/config.json` to pick up any available ECR credentials; they Just Work™️.

Except...

Through this usage, this library logs a harmless error when we check whether a particular possibly-non-ECR registry can be used with this code. In its normal CLI usage, `docker` won't invoke the ecr-login cred helper unless it already knows the registry is ECR, so this logging is correct in identifying a weird situation. But in usage as a library, this leads to spammy, sometimes scary log messages, which we'd like to avoid.

I tried all manner of terrible hacks to disable `os.Stderr`, but due to how logrus works internally, and how this code uses it, it doesn't seem possible without making logrus-specific changes to all those downstream clients.

Instead, this PR proposes adding a func that's unused by the CLI code (its logic remains unchanged), to let library consumers set the log output for the scope of ecr-login's usage, which downstream clients can set to `ioutil.Discard` if they want. This is quite a bit less invasive than any alternative I came up with, and should impose almost no maintenance burden to you lovely maintainers ❤️ 

This also slightly changes the API for instantiating an `ECRHelper` using `NewECRHelper()`, though AFAIK the CLI (and our users, whom I'll happily update) depends on that directly.

### DIFF
--- a/ecr-login/ecr.go
+++ b/ecr-login/ecr.go
@@ -16,6 +16,7 @@ package ecr
 import (
 	"errors"
 	"fmt"
+	"io"
 
 	"github.com/sirupsen/logrus"
 
@@ -26,7 +27,34 @@ import (
 var notImplemented = errors.New("not implemented")
 
 type ECRHelper struct {
-	ClientFactory api.ClientFactory
+	clientFactory api.ClientFactory
+	logger        *logrus.Logger
+}
+
+type Option func(*ECRHelper)
+
+func WithClientFactory(clientFactory api.ClientFactory) Option {
+	return func(e *ECRHelper) {
+		e.clientFactory = clientFactory
+	}
+}
+
+func WithLogOutput(w io.Writer) Option {
+	return func(e *ECRHelper) {
+		e.logger.Out = w
+	}
+}
+
+func NewECRHelper(opts ...Option) *ECRHelper {
+	e := &ECRHelper{
+		clientFactory: api.DefaultClientFactory{},
+		logger:        logrus.StandardLogger(),
+	}
+	for _, o := range opts {
+		o(e)
+	}
+
+	return e
 }
 
 // ensure ECRHelper adheres to the credentials.Helper interface
@@ -45,7 +73,7 @@ func (ECRHelper) Delete(serverURL string) error {
 func (self ECRHelper) Get(serverURL string) (string, string, error) {
 	registry, err := api.ExtractRegistry(serverURL)
 	if err != nil {
-		logrus.
+		self.logger.
 			WithError(err).
 			WithField("serverURL", serverURL).
 			Error("Error parsing the serverURL")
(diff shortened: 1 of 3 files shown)


ecr-login/ecr.go
```diff
@@ -26,7 +27,20 @@ import (
 var notImplemented = errors.New("not implemented")
 
 type ECRHelper struct {
-	ClientFactory api.ClientFactory
+	clientFactory api.ClientFactory
+	logger        *logrus.Logger
```
kzys, line 31: How about using FieldLogger?

ecr-login/ecr.go
```diff
@@ -26,7 +27,20 @@ import (
...
+}
+
+func NewECRHelper(cf api.ClientFactory) ECRHelper {
+	return ECRHelper{
+		clientFactory: cf,
+		logger:        logrus.StandardLogger(),
+	}
+}
+
+func (self ECRHelper) WithLogOutput(w io.Writer) ECRHelper {
```
kzys, line 41: I'm inclined to use FieldLogger here as well. The caller may want to customize the logger (e.g. use JSON logging)

imjasonh: This would tie ecr-login to logrus as its logging implementation via its exported API. I'm not sure how important that flexibility is, but currently, and with this change, ecr-login could change to another logging library under the hood, without callers being aware or needing to make any changes.

AFAIK, if this only supported `WithFieldLogger`, a caller that wanted to disable ecr-login's logging entirely would have to implement a discarding `FieldLogger` implementation, or else take a dependency on logrus and call `logrus.New().SetOutput(ioutil.Discard)`.

If you think it would be a useful addition as a separate method, I'm happy to add it though.

ecr-login/ecr.go
imjasonh: Does this give any flexibility to redirect/discard the output? I'm not very familiar with logrus personally, but it seems to only support log-writing and field-setting methods: https://pkg.go.dev/github.com/sirupsen/logrus#FieldLogger

kzys: You can either create a logrus.Logger and then set its Out as you did, or use https://github.com/sirupsen/logrus/blob/fdf1618bf7436ec3ee65753a6e2999c335e97221/hooks/test/test.go#L42.

imjasonh: Right. In either of those cases, my code has to be aware that logrus is involved, and take an explicit direct dependency on it.

go-containerregistry and many of its users don't depend on logrus directly today, and might only depend on it through ecr-login. Exposing that dependency through ecr-login's API means you can't change it in the future (if you want to) without breaking folks. By only exposing stdlib interfaces in the API, callers can keep a somewhat tidier dependency graph, and you retain some flexibility for the future. AFAIK nobody has expressed a desire to pass a different formatter.

imjasonh: As an example, here's how I intend to use this affordance in `ko` to disable ecr-login's log message, without having to expose logrus as an implementation detail to `ko`.

If this PR is merged, I'd also make a similar change to tools like `cosign`, Tekton, Kaniko, etc., for the same reason.

imjasonh:
> 

Yeah I can do that too if you'd like. One sec.

kzys:
Looks good! Can you squash the commits to make the history tidy? Please keep `Sign-off` line. We need to have that on all commits (yeah, we should check that in GitHub Actions...).

kzys (approved):

Issue #154, opened 2019-02-09
Don't throw exceptions from ECRHelper methods


bertramn:
There are 2 comments in the `Add` and `Delete` method implementation of the Docker [ECRHelper interface](https://github.com/awslabs/amazon-ecr-credential-helper/blob/master/ecr-login/ecr.go#L37) that these methods are not needed, followed by a throw error. Incidentally tools like packer **DO** call these methods resulting in our ECR based container builds failing.

Any chance you can log that the method was called but not throw an error? Something like this is sufficient.

```go
func (ECRHelper) Add(creds *credentials.Credentials) error {
   // This is not implemented but called by packer
   logrus.
      WithField("serverURL", creds.ServerURL).
      WithField("username", creds.Username).
      Warn("credentials store method was called which is currently not implemented")
   return nil
}
func (ECRHelper) Delete(serverURL string) error {
   // This is not implemented but called by packer
   logrus.
      WithField("serverURL", serverURL).
      Warn("credentials erase method was called which is currently not implemented")
   return nil
}
```

samuelkarp:
Hi @bertramn, thanks for opening this issue.  I'm curious to know what code is expecting to call `Add` and `Delete` when using this helper; is it code that expects the methods to actually do something or is it just boilerplate and you're really okay with it being ignored?

I'm not sure how I feel about changing these to appear successful when they're not.  In general, I'm not a fan of software that doesn't tell me when it's wrong, but I'm assuming you don't have full control over whatever is invoking the `Add` and `Delete` commands?

bertramn:
I tried to use the ECR helper with [packer](https://www.packer.io) to build docker images. When I look at the log record, it appears that Packer is calling these APIs after authenticating to delegate storing of the credentials (e.g. a mechanism you will find in the OSX keychain helper). Not an expert but the logic of calling the API makes sense, the part of how you mean to "chain" the cred helpers does not. If the Add and Delete methods fail, my packer builds fail.

julienduchesne:
If I may chime in too. Having these commands crash makes it impossible to migrate from a workflow where `$(aws ecr get-login --no-include-email)` is called to using this.

Our specific use case is this:
We want to use this for Jenkins slave AMIs but our developers are using lots of logins in their pipelines. If we install this on AMIs, we will crash everyone's pipelines. 

Maybe we shouldn't block a regular login even if we have this installed?

julienduchesne:
[...]

Issue #102, opened 2018-04-16
error storing credentials - err: exit status 1, out: `not implemented`


shumkov:
Hi there,

I got the strange error:

```bash
~/build/dashevo/dashdrive$ eval $(aws ecr get-login --no-include-email)
WARNING! Using --password via the CLI is insecure. Use --password-stdin.
Error saving credentials: error storing credentials - err: exit status 1, out: `not implemented`

~/build/dashevo/dashdrive$ cat ~/.docker/config.json
{
    "credsStore":"ecr-login"
}

~/build/dashevo/dashdrive$ docker-credential-ecr-login version
0.6.0

~/build/dashevo/dashdrive$ docker version
Client:
 Version:       18.04.0-ce
 API version:   1.37
 Go version:    go1.9.4
 Git commit:    3d479c0
 Built: Tue Apr 10 18:21:14 2018
 OS/Arch:       linux/amd64
 Experimental:  false
 Orchestrator:  swarm

Server:
 Engine:
  Version:      18.04.0-ce
  API version:  1.37 (minimum version 1.12)
  Go version:   go1.9.4
  Git commit:   3d479c0
  Built:        Tue Apr 10 18:19:24 2018
  OS/Arch:      linux/amd64
  Experimental: false
```

Credentials exported as env vars:
```
Setting environment variables from repository settings
$ export AWS_SECRET_ACCESS_KEY=[secure]
$ export AWS_ACCESS_KEY_ID=[secure]
$ export AWS_DEFAULT_REGION=[secure]
```

shumkov:
Ops... Sorry, I should not use `eval $(aws ecr get-login --no-include-email)` anymore.

apremalal:
@shumkov what did you use to solve this issue?

bondxf:
@shumkov I also experienced the same issue. Could you share what you did to solve the issue?

shumkov:
@apremalal @bondxf 
1. Configure docker daemon:
```
{
    "credsStore":"ecr-login"
}
```
2. Export your AWS credentials to env:
```
export AWS_SECRET_ACCESS_KEY=[secure]
export AWS_ACCESS_KEY_ID=[secure]
export AWS_DEFAULT_REGION=[secure]
```

Then you can pull/push your image without `docker login`.

marcomagdy:
This is how you solve the issue actually. https://github.com/aws/aws-cli/issues/3264#issuecomment-381478869

[...]

PR #315, opened 2022-03-11
Optionally ignore docker's request for storing / deleting credentials

dotboris:
*Issue #, if available:* #102 & #154

*Description of changes:*

Currently, both the `Add` and `Delete` operations in this helper return a `not implemented` error. This can lead to issues where `docker login` and `docker logout` stop working for the user while displaying a confusing `not implemented` error.

Following the suggestion in https://github.com/awslabs/amazon-ecr-credential-helper/issues/154#issuecomment-475000495, this PR adds an environment variable (`AWS_ECR_IGNORE_CREDS_STORAGE`) to optionally ignore these requests and return no error. When `AWS_ECR_IGNORE_CREDS_STORAGE=true`, instead of returning a `not implemented` error, we return `nil`.

### DIFF
--- a/ecr-login/ecr.go
+++ b/ecr-login/ecr.go
@@ -14,14 +14,18 @@
 package ecr
 
 import (
+	"errors"
 	"fmt"
 	"io"
+	"os"
 
 	"github.com/sirupsen/logrus"
 
 	"github.com/awslabs/amazon-ecr-credential-helper/ecr-login/api"
 	"github.com/docker/docker-credential-helpers/credentials"
 )
+
+var notImplemented = errors.New("not implemented")
 
 type ECRHelper struct {
 	clientFactory api.ClientFactory
@@ -67,27 +71,37 @@ func NewECRHelper(opts ...Option) *ECRHelper {
 // ensure ECRHelper adheres to the credentials.Helper interface
 var _ credentials.Helper = (*ECRHelper)(nil)
 
+func shouldIgnoreCredsStorage() bool {
+	return os.Getenv("AWS_ECR_IGNORE_CREDS_STORAGE") == "true"
+}
+
 // Called when docker tries to store credentials. This usually happens during `docker login` calls. In our context,
-// storing arbitrary user given credentials makes no sense. To keep interoperability with applications that call `docker
-// login` during their execution, this is implemented as a nop.
+// storing arbitrary user given credentials makes no sense.
 func (self ECRHelper) Add(creds *credentials.Credentials) error {
-	self.logger.
-		WithField("username", creds.Username).
-		WithField("serverURL", creds.ServerURL).
-		Warning("Ignoring request to store credentials. " +
-			"This is not supported in the context of the docker ecr-login helper.")
-	return nil
+	if shouldIgnoreCredsStorage() {
+		self.logger.
+			WithField("username", creds.Username).
+			WithField("serverURL", creds.ServerURL).
+			Warning("Ignoring request to store credentials. " +
+				"This is not supported in the context of the docker ecr-login helper.")
+		return nil
+	} else {
+		return notImplemented
+	}
 }
 
 // Called when docker tries to delete credentials. This usually happens during `docker logout` calls. In our context, we
-// don't store arbitrary user given credentials so deleting them makes no sense. To keep interoperability with
-// applications that call `docker logout` during their execution, this is implemented as a nop.
+// don't store arbitrary user given credentials so deleting them makes no sense.
 func (self ECRHelper) Delete(serverURL string) error {
-	self.logger.
-		WithField("serverURL", serverURL).
-		Warning("Ignoring request to delete credentials. " +
-			"This is not supported in the context of the docker ecr-login helper.")
-	return nil
+	if shouldIgnoreCredsStorage() {
+		self.logger.
+			WithField("serverURL", serverURL).
+			Warning("Ignoring request to delete credentials. " +
+				"This is not supported in the context of the docker ecr-login helper.")
+		return nil
+	} else {
+		return notImplemented
+	}
 }
 
 func (self ECRHelper) Get(serverURL string) (string, string, error) {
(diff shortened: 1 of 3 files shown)


kzys:
Let me check with Docker folks first.

GCP is also having the same issue (see https://github.com/GoogleCloudPlatform/docker-credential-gcr/blob/e8a16dd0fa5bb661e304ad850a61ccfaa901045d/credhelper/helper.go#L62-L75). So I'd like to have a vendor-neutral solution.

danxmoran:
Commenting to nudge this 😅

Something like this would be a big help for my team. We're trying to migrate our CI from using `docker login` to using the ECR helper. We hoped that we could do it by adding the helper (and config) to our CI image, then incrementally removing the `docker login` usage from workflows. As soon as we pushed the change configuring the helper in `~/.docker/config.json` all of our workflows started to fail on the `docker login` steps, and we had to revert.

hskrtich:
Would it be possible to add this change now and reconfigure if/when there is a central way to manage this?

jasonheffner:
I'm guessing this project is no longer being developed. The last release was over a year ago. This would have been very helpful.

a544jh:
In my opinion an environment variable shouldn't even be needed. The helper is caching the token anyway, so the `store` command could just validate your AWS credentials and cache the token. Likewise `erase` could just delete the cache. If caching is disabled these operations cold just be NOPs, and possibly print out a warning.

By throwing errors (as it's doing currently), this helper is essentially not correctly implementing the docker credential helper spec, as can be seen when numerous people are having issues with this helper when `docker login` isn't working.

junze-smg:
Is there any movement to getting this PR reviewed? We ran into this problem today that `docker login` command failed on `not implemented` when we specify `ecr-login` in the config file.

ecr-login/ecr.go
```diff
@@ -70,14 +71,37 @@ func NewECRHelper(opts ...Option) *ECRHelper {
...
-	// This does not seem to get called
-	return notImplemented
+func shouldIgnoreCredsStorage() bool {
+	return os.Getenv("AWS_ECR_IGNORE_CREDS_STORAGE") == "true"
 }
 
-func (ECRHelper) Delete(serverURL string) error {
-	// This does not seem to get called
-	return notImplemented
+// Called when docker tries to store credentials. This usually happens during `docker login` calls. In our context,
```
ginglis13, line 78: nit: godoc 
```suggestion
// Add tries to store credentials when docker requests it. This usually happens during `docker login` calls. In our context,
```

ecr-login/ecr.go
```diff
@@ -70,14 +71,37 @@ func NewECRHelper(opts ...Option) *ECRHelper {
...
+			WithField("serverURL", creds.ServerURL).
+			Warning("Ignoring request to store credentials. " +
+				"This is not supported in the context of the docker ecr-login helper.")
+		return nil
+	} else {
+		return notImplemented
+	}
+}
+
+// Called when docker tries to delete credentials. This usually happens during `docker logout` calls. In our context, we
```
ginglis13, line 93: nit: godoc 
```suggestion
// Delete tries to delete credentials when docker requests it. This usually happens during `docker logout` calls. In our context, we
```

ginglis13 (approved):
godoc nits otherwise LGTM

ecr-login/ecr.go
```diff
@@ -70,14 +71,37 @@ func NewECRHelper(opts ...Option) *ECRHelper {
...
-	return notImplemented
+// Add tries to store credentials when docker requests it. This usually happens during `docker login` calls. In our context,
+// storing arbitrary user given credentials makes no sense.
+func (self ECRHelper) Add(creds *credentials.Credentials) error {
+	if shouldIgnoreCredsStorage() {
+		self.logger.
+			WithField("username", creds.Username).
+			WithField("serverURL", creds.ServerURL).
+			Warning("Ignoring request to store credentials. " +
+				"This is not supported in the context of the docker ecr-login helper.")
```
swagatbora90, line 86: Wouldn't it be more useful to output this error message when `shouldIgnoreCredsStorage()` is not true? and add something like

```
To ignore this error, set env value `AWS_ECR_IGNORE_CREDS_STORAGE` to true
```

webratz:
Is there any hope that this will see some traction and get merged?

swagatbora90:
@dotboris Are you still working on this? Looks like there is a lot of interest to get this feature in.

dotboris:
I'm not. I haven't touched this in ages. I don't really have the bandwidth to work on this.

jamestelfer:
@swagatbora90 I can pick this up if you're able to review?

ecr-login/ecr.go
therealvio: We're interested in getting this change merged. I am picking up this PR through here: https://github.com/awslabs/amazon-ecr-credential-helper/pull/847 so we can get feedback addressed and hopefully merged soon. Can you please relay feedback for this PR there and I can look at resolving these.

In the meantime:
Can you elaborate on what you mean @swagatbora90? I am not quite sure on what the intention of this suggestion is.

therealvio: Upon thinking about it for longer, I think I understand. I will implement this in my draft.

therealvio: My understanding is that we're preserving the `notImplemented` error, and this particular behaviour is what users want to avoid when all they want to do is authenticate, whether or not storing the credential is supported doesn't matter. They want to get the auth done.

The environment variable is meant to provide a means to skip the attempt at storing the credential and throw a non-blocking warning rather than an error. If we inverted the condition for checking `shouldIgnoreCredsStorage`, users will still get the `notImplemented` error and doesn't satisfy what the change was intending to do. The test `TestAddIgnored` reflects this behaviour.

therealvio:
This PR can be closed in favour of #847 which has picked up the work and finished it off.

therealvio:
Update for onlookers: The changes proposed in here, which were completed in #847 have now been merged in. Enjoy.

PR #326, opened 2022-06-22
Don't require amazonaws.com in ECR url pattern

jtwalraven:
*Issue #, if available:*

N/A

*Description of changes:*

Some deployments of AWS are served behind a proxy or have an alternative domain name. The current code checks for `amazonaws.com` in the domain name when it gets matches via the regex (see https://github.com/awslabs/amazon-ecr-credential-helper/blob/main/ecr-login/api/client.go#L40). The domain name does not appear to be used in the matches argument and could be removed to enable using this against proxy, alternative, or iso region deployments of AWS.

### DIFF
--- a/ecr-login/api/client.go
+++ b/ecr-login/api/client.go
@@ -37,7 +37,7 @@ const (
 	ecrPublicEndpoint   = proxyEndpointScheme + ecrPublicName
 )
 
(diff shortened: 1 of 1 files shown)


supergibbs:
Custom DNS for ECR isn't supported and if you did setup a custom proxy, why would you still use all the stuff before `amazon.com`. Wouldn't you just use something like `ecr.mycompany.com`?

jtwalraven:
Resolved by #434

PR #400, opened 2022-09-09
Fix Missing build-id issue during rpm build

swagatbora90:
Pass --buildid as ldflag otherwise rpmbuild fails. Golang compiler adds its own build-id into the binary, named .note.go.build-id. However, rpmbuild wants Fedora compatible build-id, in the format note.gnu.build-id.

Signed-off-by: Swagat Bora <sbora@amazon.com>

*Issue #, if available:* https://t.corp.amazon.com/P68468819

*Description of changes:*
Saw that the `main` branch is already passing the `buildid` ldflag, which was missing in the `amazonlinux` branch. This was causing rpmbuild to fail. There was another issue which I believe is caused due to stripping on symbol table while the binaries were built. This was fixed by passing -D 'debug_package %{nil}` during rpmbuild

```
RPM build errors:
    Empty %files file /home/ec2-user/amazon-ecr-credential-helper/BUILD/amazon-ecr-credential-helper-0.6.0/debugsourcefiles.list
```

### DIFF
--- a/scripts/build_binary.sh
+++ b/scripts/build_binary.sh
@@ -35,6 +35,10 @@ if [[ -n "${3}" ]]; then
 fi
 
 GOOS=${TARGET_GOOS:-} GOARCH=${TARGET_GOARCH:-} CGO_ENABLED=0 \
-       	go build -installsuffix cgo -a -ldflags "-s ${version_ldflags}" \
-       	-o ../$1/docker-credential-ecr-login \
-	./cli/docker-credential-ecr-login
+        go build \
+        -installsuffix cgo \
+        -a \
+        -ldflags "-buildid= -s ${version_ldflags}" \
+        -trimpath \
+        -o ../$1/docker-credential-ecr-login \
+        ./cli/docker-credential-ecr-login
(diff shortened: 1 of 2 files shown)


austinvazquez (approved):

ginglis13 (approved):

scripts/build_binary.sh
```diff
@@ -35,6 +35,10 @@ if [[ -n "${3}" ]]; then
 fi
 
 GOOS=${TARGET_GOOS:-} GOARCH=${TARGET_GOARCH:-} CGO_ENABLED=0 \
-       	go build -installsuffix cgo -a -ldflags "-s ${version_ldflags}" \
-       	-o ../$1/docker-credential-ecr-login \
-	./cli/docker-credential-ecr-login
+        go build \
+        -installsuffix cgo \
+        -a \
+        -ldflags "-buildid= -s ${version_ldflags}" \
```
swagatbora90, line 41: Explicitly adding `-buildid=` removes any buildid from being added to the binary, which fixes the issue of go build-id style mismatch. I don't think we have any use for build-id, so think this will work. If my understanding is wrong, we can also explicitly pass a GNU style buildid using `-B` option.

PR #502, opened 2023-02-25
Add dependabot auto-merge

austinvazquez:
*Issue #, if available:*
N/A

*Description of changes:*
Add auto-merge of dependabot pull-requests.

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -1,0 +1,28 @@
+name: Dependabot auto-merge
+on: pull_request
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  dependabot:
+    runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
+    steps:
+      - name: Dependabot metadata
+        id: metadata
+        uses: dependabot/fetch-metadata@v1
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Approve and auto-merge PR
+        if: |
+          steps.metadata.outputs.update-type == 'version-update:semver-minor' || 
+          steps.metadata.outputs.update-type == 'version-update:semver-patch'
+        run: |
+          gh pr review --approve "$PR_URL"
+          gh pr merge --auto --merge "$PR_URL"
+        env:
+          PR_URL: ${{github.event.pull_request.html_url}}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}


.github/workflows/automerge.yml
```diff
@@ -0,0 +1,29 @@
...
+        with:
+          github-token: "${{ secrets.GITHUB_TOKEN }}"
+
+      - name: Approve a PR
+        run: gh pr review --approve "$PR_URL"
+        env:
+          PR_URL: ${{github.event.pull_request.html_url}}
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Enable auto-merge for Dependabot PRs
```
swagatbora90, line 25: Should we protect the `amazon-ecr-credential-helper` main branch first, before enabling this workflow? We should make sure that the status checks pass before a patch gets merged.

swagatbora90 (changes requested):

.github/workflows/automerge.yml
austinvazquez: Yeah sure, my initial impression from my research was this only merge after status checks, but let me verify that behavior. I might just need to add a sync step between the workflows. Something like if all checks pass, approve and merge.

austinvazquez: Oh I understand now after a second read of https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions#enable-auto-merge-on-a-pull-request

I'll look at our settings.

austinvazquez: @swagatbora90, I updated branch protections to require status check pass before merging and require one approval.

swagatbora90: Also, should we limit this to only minor or patch version upgrades?  Major version upgrade can introduce  breaking changes, so it may be better to have some manual review. The above link does something similar.

```
- name: Enable auto-merge for Dependabot PRs
        if: steps.metadata.outputs.update-type == 'version-update:semver-patch'}}
        run: gh pr merge --auto --merge "$PR_URL"
        env:
          PR_URL: ${{github.event.pull_request.html_url}}
          GITHUB_TOKEN: ${{secrets.GITHUB_TOKEN}}
```

swagatbora90 (commented):
LGTM.

.github/workflows/automerge.yml
austinvazquez: That makes sense to me. Let's start with just minor and patch and see if we can reduce some of the manual review. Off the top of my head, AWS SDK pushes lots of minor and patch updates, but like you mention a major update would probably be best for us to take a look.

swagatbora90 (approved):
LGTM!

ginglis13 (approved):

PR #504, opened 2023-03-12
Allow users to enable the default registry

amancevice:
*Issue #, if available:*

N/A

*Description of changes:*

While not _officially_ supported, it is possible to alias an ECR registry with DNS (eg ecr.example.io).

This change adds support for the `AWS_ECR_USE_DEFAULT_REGISTRY` environment variable which, when set to a non-empty value, will force this tool to use the default registry when authenticating.

I am not married to this implementation; if the maintainers have a better idea for allowing users to use this tool with non-AWS-style registry names I'm happy to try to implement!


rblaine95:
Stumbled into this while working on trying to get https://github.com/containerd/containerd/discussions/11385 working...

This would be a very, very nice feature to have.

Some way to tell the ECR Credential Helper that, hey, if you see (e.g) `docker.io`, then load up the credentials for some specific ECR registry.

amancevice:
I'm going to close this for lack of response from the AWS Labs team. I'm pretty disappointed in the lack of guidance here; happy to reopen if someone actually gets back to me!

PR #513, opened 2023-03-22
Modify automerge to only run after build and scan workflows

austinvazquez:
*Issue #, if available:*
N/A

*Description of changes:*
Automerge is failing on dependabot pull requests because the workflow is running before build and security scan workflows are complete.

### DIFF
--- a/.github/workflows/auto-merge.yml
+++ b/.github/workflows/auto-merge.yml
@@ -1,5 +1,10 @@
 name: Dependabot auto-merge
-on: pull_request
+on:
+  pull_request:
+    branches: [ "main" ]
+  workflow_run:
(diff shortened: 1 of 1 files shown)


austinvazquez:
After some more research into GitHub Actions workflow scheduling I found the workflow triggers currently have too many limitations. Reference https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

1. Workflow will run if any of the dependency workflow completes.
2. Workflows can be chained, but is limited to a depth of 3. So while Build -> Security Scan -> Automerge might be feasible right now, if we wanted to add a release workflow then that falls apart.

Issue #574, opened 2023-07-22
Getting the error error storing credentials - err: exit status 1, out: `not implemented`


dean-hpe:
I'm running a Ubuntu22 VM using VMWare Fusion 13.  The host OS is MacOS on M1 Apple silicon.  I build docker-credential-ecr-login following the docs like this:

`go install github.com/awslabs/amazon-ecr-credential-helper/ecr-login/cli/docker-credential-ecr-login@latest`

The following command gives this error:

```
aws ecr get-login-password --region us-west-2 | docker login --username AWS --password-stdin <ID>.dkr.ecr.us-west-2.amazonaws.com
Error saving credentials: error storing credentials - err: exit status 1, out: `not implemented`
```
Lots of people have seen this and none of the suggestions have worked.

merlinran:
The culprit is https://github.com/awslabs/amazon-ecr-credential-helper/blob/main/ecr-login/ecr.go#L75 . Just changing both methods to return `nil` fixed the error. Will have a PR soon

pauldthomson:
I don't know if that's the correct fix, rather shouldn't it be doing what [this line](https://github.com/awslabs/amazon-ecr-credential-helper/blob/5c087f35e29c21d4e42a4c999c7e6c4cad32d7db/ecr-login/api/client.go#L249) is doing? 
That function is called by `docker login` which you don't technically need to use with this helper, but I'm not sure if that's breaking the [credential helper protocol](https://docs.docker.com/engine/reference/commandline/login/#credential-helper-protocol)?

jaswanthikolla:
We are facing this issue too. When you have a common Github action runner with different use cases, there are users who uses this [ECR login aws action ](https://github.com/aws-actions/amazon-ecr-login). This action basically does `docker login` and gives the error `exit status 1, out: not implemented` ( Note that ecr-credential-helper is installed) 

Any update on this?

jaswanthikolla:
We are facing this issue too. When you have a common Github action runner with different use cases, there are users who uses this [ECR login aws action ](https://github.com/aws-actions/amazon-ecr-login). This action basically does `docker login` and gives the error `exit status 1, out: not implemented` ( Note that ecr-credential-helper is installed) 

Any update on this issue?

Workaround for this issue:
```
{
    "credHelpers": {
		"xxx.dkr.ecr.us-east-2.amazonaws.com": "ecr-login"
	}
}
```

pauldthomson:
[...]

Issue #474, opened 2023-01-26
Using this credential helper prevents you from pulling from docker hub


sterling000:
pretty easy to reproduce
follow the setup instructions
configure with the credsStore ecr-login (the credHelpers never worked for me)
pull an image from your private repo
then call docker login

Observe Error:
Error saving credentials: error storing credentials - err: exit status 1, out: `not implemented`

ribeirofael:
Same problem here.

niroliyanage:
same issue here had to override  the DOCKER_CONFIG='~./docker/bla.json'  to get it to work with other repositories

xocasdashdash:
@hannes-ucsc are you able to still do:
```
 aws --region us-east-1 ecr get-login-password | docker login --username AWS --password-stdin XXXX.dkr.ecr.us-east-1.amazonaws.com
 ```
 with that config? because i have that exact config and it's failing with 
 ```
 Error saving credentials: error storing credentials - err: exit status 1, out: `not implemented`
 ```

hannes-ucsc:
I don't know. I don't use `docker login` in contexts where the helper is installed. I thought the point of this helper is that one doesn't need to use `docker login`.

Issue #316, opened 2022-03-15
After upgrading to macos 12.3 I'm getting "Error saving credentials: error storing credentials - err: exit status 1, out: `not implemented`"


wilgucki:
Everything was working fine before the upgrade but once I've upgraded to macos 12.3 I'm getting this error
`Error saving credentials: error storing credentials - err: exit status 1, out: not implemented`
when trying to run this
`aws ecr get-login-password --region xxx | docker login --username AWS --password-stdin xxx.dkr.ecr.xxx.amazonaws.com`

Running `docker-compose up` fails with this error
`Error response from daemon: pull access denied for xxx, repository does not exist or may require 'docker login': denied: requested access to the resource is denied`

I've double checked access rights - all seems to be fine (plus nothing has changed recently in this area).

Here's my docker config
```
~ cat ~/.docker/config.json --plain
{
	"credHelpers": {
		"xxx.dkr.ecr.xxx.amazonaws.com": "ecr-login"
	}

}
```

I have verified that everything is fine on macos 12.2 (asked few people to run mentioned commands). I also verified with other person running macos 12.3 that this doesn't work. I have tested this on docker desktop 4.5 and 4.6.

wilgucki:
I think I found workaround for my issue
```
rm /usr/local/bin/docker-credential-ecr-login
brew uninstall docker-credential-helper-ecr
brew cleanup
ln -s /Applications/Docker.app/Contents/Resources/bin/docker-credential-ecr-login /usr/local/bin/docker-credential-ecr-login
```

To make containers work I need to do this

```
docker-compose build
docker-compose pull
docker-compose up -d
```

Sometimes I was able to use a single command instead (`docker-compose up --build -d`) but in most cases I'm getting this error
```
Error response from daemon: pull access denied for <some name>, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```
So, reinstalling credentials helper and using separate commands to deal with containers seems to be a good enough workaround but the main issue has not been fixed.

timfallmk:
I am also seeing this issue while using SSO. The binaries released via homebrew and Docker seem to be different.
```shell
diff /opt/homebrew/Cellar/docker-credential-helper-ecr/0.6.0/bin/docker-credential-ecr-login /Applications/Docker.app/Contents/Resources/bin/docker-credential-ecr-login
[...]

PR #575, opened 2023-07-23
fix `not implemented` errors

merlinran:
*Issue #, if available:*
Closes https://github.com/awslabs/amazon-ecr-credential-helper/issues/574
Closes https://github.com/awslabs/amazon-ecr-credential-helper/issues/474
Closes https://github.com/awslabs/amazon-ecr-credential-helper/issues/316

*Description of changes:*
The current code assumes some methods are not called, which is no longer true.

### DIFF
--- a/ecr-login/ecr.go
+++ b/ecr-login/ecr.go
@@ -71,13 +71,13 @@ func NewECRHelper(opts ...Option) *ECRHelper {
 var _ credentials.Helper = (*ECRHelper)(nil)
 
 func (ECRHelper) Add(creds *credentials.Credentials) error {
-	// This does not seem to get called
-	return notImplemented
+	// Never store the credentials
+	return nil
 }
 
 func (ECRHelper) Delete(serverURL string) error {
-	// This does not seem to get called
-	return notImplemented
+	// Never store the credentials
+	return nil
 }
 
 func (self ECRHelper) Get(serverURL string) (string, string, error) {


sparr (approved):

sparr:
Simply returning no error when a caller would expect functionality to be triggered does not seem appropriate. Can you elaborate on the current use cases for these methods, to make review smoother?

ginglis13:
> Simply returning no error when a caller would expect functionality to be triggered does not seem appropriate. Can you elaborate on the current use cases for these methods, to make review smoother?

Agreed. The ECR credential helper is intended to be used *instead* of `login` and `logout`. I think silently returning no error is not enough here, I'd suggest keeping the error return and adding a warning log statement that this tool is intended to be set up as a credential helper in `~/.docker/config.json` (or `~/.finch/config.json` if using [`finch`](https://github.com/runfinch/finch))

See https://github.com/awslabs/amazon-ecr-credential-helper#troubleshooting and https://github.com/awslabs/amazon-ecr-credential-helper#usage

merlinran:
Sorry I thought it's what prevented my `ecr-login` from working. It turns out to be a different issue. I noticed the error message during troubleshooting, then came to the conclusion too quickly.

As a side note, for anyone who by chance lands here, what prevented `ecr-login` from working for docker 24.0.4 with `containerd-snapshotter` enabled was because of the following:
```
time="2023-07-23T21:28:08.336275463Z" level=warning msg="Host doesn't match" cfgHost=registry-1.docker.io host=xxx.amazonaws.com
time="2023-07-23T21:28:08.342797167Z" level=error msg="Handler for POST /v1.43/images/create returned error: failed to resolve reference \"xxx.amazonaws.com/<image>:latest\": pulling from host xxx.amazonaws.com failed with status code [manifests latest]: 401 Unauthorized"
```
and what fixed it for me was with this in `~/.docker/config.json`:
```
{
	"auths": {
		"xxx.amazonaws.com": {}  <=== has to add this to avoid the "Host doesn't match" warning
	},
	"credHelpers": {
		"xxx.amazonaws.com": "ecr-login"
	}
}
```

PR #792, opened 2024-04-24
Adds back Dockerfile for local builds

austinvazquez:
*Issue #, if available:*
The Dockerfile for building the credential helper was removed in #784 to simplify maintenance. The issue is the container is helpful for users which need to build the credential helper from source with only Docker as a dependency.

*Description of changes:*
This commit reverts the removal of the repositories Dockerfile for local builds. The container was helpful for users which need to build the credential helper from source but do not have an environment to do so.

In addition to the revert, this chance adds dependabot configuration to update the base image for the container and reworks the docker make target to build artifacts with correct user permissions.

*How was it tested?*
Ran `make build-in-docker` and `make all-variants-in-docker` in Amazon Linux 2023 EC2 instance with only Docker installed. Also verified the artifacts generated have correct user permissions. i.e. current user instead of root

Merged change to fork and verified dependabot automation opened [PR](https://github.com/austinvazquez/amazon-ecr-credential-helper/pull/67) for updating base container.

Cross-compilation using container in CI is successful.


Shubhranshu153 (approved):

pendo324 (approved):
Reasoning makes sense and implementation looks good. Nice wildcard target

PR #794, opened 2024-04-26
Add disclaimer for container image support to readme

austinvazquez:
*Issue #, if available:*
Clarify the support model for the Dockerfile included in the Amazon ECR credential helper repository.

*Description of changes:*
This change adds a disclaimer regarding the support of the Dockerfile provided for cross compiling the Amazon ECR credential helper binaries and locally building from source using Docker.

### DIFF
--- a/README.md
+++ b/README.md
@@ -163,6 +163,11 @@ To install run:
 go install github.com/awslabs/amazon-ecr-credential-helper/ecr-login/cli/docker-credential-ecr-login@latest
 ```
 
+> [!WARNING]
+> Disclaimer: the [Dockerfile](./Dockerfile) in this repository is used to test cross-compilation of the
+> Amazon ECR credential helper binaries in GitHub Actions CI and as a developer utility for building locally from source.
+> It is a reference implementation and not security hardened for building and running production containers.
+
 If you already have Docker environment, just clone this repository anywhere
 and run `make build-in-docker`. This command builds the binary with Go inside the Docker
 container and output it to local directory.


coderbirju (approved):

README.md
```diff
@@ -163,6 +163,11 @@ To install run:
 go install github.com/awslabs/amazon-ecr-credential-helper/ecr-login/cli/docker-credential-ecr-login@latest
 ```
 
+> [!WARNING]
+> Disclaimer: the container image provided here is to validate cross-compilation of the Amazon ECR credential
+> helper binaries in GitHub Actions CI and as a reference implementation for use cases which require
+> building from source. It is **not** recommended to run this container image in production environments.
```
pendo324, line 169: A couple nits here.

1. We're not really providing a container image, we're just publishing a Dockefile right? Like, we aren't ever pushing to a registry are we?
1. What does "run the container image in production" even mean in the context of our Dockerfile? It just installs some build time deps and runs a `make` command, right? I guess the only way to use it in production would be to use our Dockerfile as a `from` and add a new run command that forwards to built ecr-credential-helper binary? Or through `docker run <our image> /path/to/ecr-ch <command>`?

austinvazquez: For 1, correct we do not publish an image to a container registry. It is just the Dockerfile in the source code repository.
For 2, perhaps it can be word smithed better, but wanted to clearly communicate the image is provided as a developer utility and reference implementation not security hardened for production usage.

pendo324 (approved):

pendo324 (approved):

PR #799, opened 2024-05-01
Add generate third party licenses

austinvazquez:
*Issue #, if available:*
The repositories' THIRD-PARTY-LICENSE file is out of date with the list of dependencies used.

*Description of changes:*
This change imports third party licenses file generation from the SOCI project and modifies it based on the Amazon ECR credential helper project's needs. Additionally this change will add license checks such that only pre-approved licenses are used in the project. This will be checked in GitHub Actions CI.


scripts/build_third_party_licenses.sh
```diff
@@ -0,0 +1,50 @@
...
+    echo "popd"
+    echo "---------------------------------------------------------------------"
+    exit 1
+fi
+
+# Remove content from the license file
+truncate -s 0 "${license_file}"
+{
+    # The apache 2.0 license doesn't get modified with a copywrite. To reduce duplication, add attribution for each project using the license, but include the license text just once.
+    go-licenses report --template="${root}/scripts/third_party_licenses/apache.tpl" --ignore github.com/awslabs/amazon-ecr-credential-helper/ecr-login ./...
```
pendo324, line 46: Some projects might have dual/multi licenses, with one being Apache 2.0. This approach would avoid copying those extra licenses, but I guess we don't really care as long as one is Apache 2.0 that's fine.

scripts/third_party_licenses/apache.tpl
```diff
@@ -0,0 +1,5 @@
+{{range . -}}
+{{if eq .LicenseName "Apache-2.0" -}}
+** {{.Name}}; version {{.Version}} - https://{{.Name}}
```
pendo324, line 3: Is every name really likely to be a valid `https://` domain?

scripts/third_party_licenses/other.tpl
```diff
@@ -0,0 +1,9 @@
+{{ range . -}}
+{{ if ne .LicenseName "Apache-2.0" -}}
+--------------------------------------------------------------------------------
+** {{.Name}}; version {{.Version}} - https://{{.Name}}
+
+{{ .LicenseText }}
```
pendo324, line 6: Can't this be one template file with the `{{ .LicenseText }}` being conditional on if its Apache 2.0 or not?

Kern--: The way this ends up in the license file is:
```
- Proj1 - info
- Proj2 - info
- Proj3 - info
Apache 2.0 license text
---
- Proj4 - info
Proj4 license text
---
- Proj5 - info
Proj5 license text
```

I couldn't figure out a way to do that with 1 template.

scripts/build_third_party_licenses.sh
Kern--: We might want to double check this, but I think the project would just show up in the list twice. Once under the Apache section, and a second time with its other license.

scripts/third_party_licenses/apache.tpl
austinvazquez: Good catch. The few I checked resolved to the Go packages page, but not every package would resolve. Instead this should be `{{.LicenseURL}}`

austinvazquez: Pushed with latest. PTAL.

scripts/build_third_party_licenses.sh
austinvazquez: It might be a limited of the go-licenses tool. See https://github.com/google/go-licenses/issues/117

austinvazquez:
@pendo324, @Kern--, please review. I was unable to solve the dual license problem. The workaround I am proposing is to include a license check in GitHub Actions CI which will fail if a license is not pre-approved. Of our existing dependencies, 1 in 26 is dual licensed. See THIRD-PARTY-LICENSES for gopkg.in/yaml.v3 (a test dependency). Both licenses are being listed in the third party licenses file currently.

pendo324 (approved):

austinvazquez:
Note to maintainers: this change was drafted such that each change is buildable. Commits in this PR _**should not**_ be squashed to preserve import of code from awslabs/soci-snapshotter project.

Issue #800, opened 2024-05-01
fatal: detected dubious ownership in repository at '/go/src/github.com/awslabs/amazon-ecr-credential-helper'


ktasper:
When running the following locally:

```
sudo -E make docker
```

I get the following error:
```
docker run --rm \
        --user 0:0 \
        --env TARGET_GOOS= \
        --env TARGET_GOARCH= \
        --volume /tmp/amazon-ecr-credential-helper:/go/src/github.com/awslabs/amazon-ecr-credential-helper \
        sha256:23b97de4f34d4c956e315659157439acfae03f15977d45c57c0e797729064846 \
        make build
git rev-parse --short=7 HEAD > GITCOMMIT_SHA
fatal: detected dubious ownership in repository at '/go/src/github.com/awslabs/amazon-ecr-credential-helper'
To add an exception for this directory, call:

        git config --global --add safe.directory /go/src/github.com/awslabs/amazon-ecr-credential-helper
make: *** [Makefile:95: GITCOMMIT_SHA] Error 128
```

However running it again seems to "fix" this problem.
```
docker run --rm \
        --user 0:0 \
        --env TARGET_GOOS= \
        --env TARGET_GOARCH= \
        --volume /tmp/amazon-ecr-credential-helper:/go/src/github.com/awslabs/amazon-ecr-credential-helper \
        sha256:23b97de4f34d4c956e315659157439acfae03f15977d45c57c0e797729064846 \
        make build
./scripts/build_binary.sh /go/src/github.com/awslabs/amazon-ecr-credential-helper/bin/local 0.7.0 
./scripts/build_binary.sh: line 33: 3: unbound variable
make: *** [Makefile:52: /go/src/github.com/awslabs/amazon-ecr-credential-helper/bin/local/docker-credential-ecr-login] Error 1
make: *** [Makefile:40: build-in-docker] Error 2
```

I have also verified this behaviour on a CICD system. 

However I do not see any docs / issues demonstrating this behaviour and it seems incorrect.

ktasper:
Fix can be found here: https://github.com/awslabs/amazon-ecr-credential-helper/pull/802

austinvazquez:
Thanks for opening and looking into this. Closing as resolved.

PR #802, opened 2024-05-02
Add awslabs to git safe dir

ktasper:
*Issue #, if available:*
Fixes the following error when running `make build-in-docker` for ISSUE: https://github.com/awslabs/amazon-ecr-credential-helper/issues/800
```
git rev-parse --short=7 HEAD > GITCOMMIT_SHA
fatal: detected dubious ownership in repository at '/go/src/github.com/awslabs/amazon-ecr-credential-helper'
To add an exception for this directory, call:

        git config --global --add safe.directory /go/src/github.com/awslabs/amazon-ecr-credential-helper
make: *** [Makefile:95: GITCOMMIT_SHA] Error 128
```

*Description of changes:*
- Adds `/go/src/github.com/awslabs/amazon-ecr-credential-helper` to git as a safe directory. 

### DIFF
--- a/scripts/container_init.sh
+++ b/scripts/container_init.sh
@@ -27,4 +27,5 @@ apk add --no-cache \
 # Resolves permission issues for Go cache when
 # building credential helper as non-root user.
 mkdir /.cache && chmod 777 /.cache
+git config --global --add safe.directory /go/src/github.com/awslabs/amazon-ecr-credential-helper
 
--- a/scripts/container_init.sh
+++ b/scripts/container_init.sh
@@ -27,5 +27,7 @@ apk add --no-cache \
 # Resolves permission issues for Go cache when
 # building credential helper as non-root user.
 mkdir /.cache && chmod 777 /.cache
+# Resolves dubious ownership of git directory when
+# building credential helper as root user.
 git config --global --add safe.directory /go/src/github.com/awslabs/amazon-ecr-credential-helper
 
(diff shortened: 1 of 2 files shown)


scripts/container_init.sh
```diff
@@ -27,4 +27,5 @@ apk add --no-cache \
 # Resolves permission issues for Go cache when
 # building credential helper as non-root user.
 mkdir /.cache && chmod 777 /.cache
+git config --global --add safe.directory /go/src/github.com/awslabs/amazon-ecr-credential-helper
```
austinvazquez, line 30: Can you add a comment like the following?

```
# Resolves dubious ownership of git directory when 
# building credential helper as root user.
```

austinvazquez (changes requested):
Thanks for looking into the issue. The challenge really comes from the build mounting `.git` directory into the container which is fine when building with non-root users but breaks as you have found when running as root users. I'd like to look more into simplifying this build logic as a seperate issue but first it would be good to get a workaround for this.

Can you apply the following patch to the Makefile along with your change? This way we generate the required shasum before going into the container as a workaround

```
diff --git a/Makefile b/Makefile
index 6fbed4f..fac7fb3 100644
--- a/Makefile
+++ b/Makefile
@@ -36,7 +36,7 @@ WINDOWS_ARM64_BINARY=$(BINPATH)/windows-arm64/$(BINARY_NAME).exe
 .PHONY: docker
 docker: build-in-docker
 
-%-in-docker:
+%-in-docker: GITCOMMIT_SHA
        docker run --rm \
                --user $(UID):$(GID) \
                --env TARGET_GOOS=$(TARGET_GOOS) \
```

austinvazquez (approved):

pendo324 (approved):
LGTM, thanks!

Issue #94, opened 2018-01-31
Support for providing registry region and id from different source


ojundt:
Hi there,

We are using ECR in combination with a proxy with a custom domain (e.g. `registry.foobar.com`). We use the custom domain to make migration from and to ECR easier. 

However, the custom domain currently prevents us from using the ECR credential helper. It requires the ECR domain format for extracting the registry id and region:
https://github.com/awslabs/amazon-ecr-credential-helper/blob/3eac0af2021820739ca704166494c1f45a4e8f72/ecr-login/api/client.go#L33-L55

It would be nice if there are alternative means to provide the region and account id information, maybe with environment variables or config files for the helper with a simple mapping such as:
```
{  
   "registry.foobar.com":{  
      "region":"eu-central-1",
      "id":"0176235112345"
   },
   "images.somethingelse.net":{  
      "region":"us-west-1",
      "id":"0182738917123"
   }
}
```

beamerblvd:
This hit me, too. It's a big bummer. I was looking forward to shortening the `<horrendous ID>.dkr.ecr.<region name>.amazonaws.com` URIs. I, apparently, successfully got a proxy in front of it, and then had to throw it all away when I had no way to authenticate (neither `ecr-login` nor `aws ecr get-login` returns an auth that Docker can use with my custom URL). Would really love to see this fixed. In my opinion (which, of course, is just that—an opinion), this is a defect, not a feature. The internet is simply full of people trying to shorten these ridiculous URIs with one proxy or another, usually with success, but it's all moot if the login tools intentionally prevent us from doing so. :-(

alexmac:
I'm surprised there isn't an easier way to front ECR behind a CNAME, it's definitely a usability failure in my opinion. Nevertheless I was able to cobble together workaround which others hitting this might be interested in.

First you need to setup some kind of proxy that can rewrite both request and response headers. I used haproxy. The relevant part of the config that reqwrites both the Host(request) and Location(Response) headers looks like this:
```
frontend fe_router
    bind 0.0.0.0:9550 ssl crt /tmp/cert.pem
    option forwardfor
    use_backend be_external_host


backend be_external_host
    http-request set-header Host 1234567890.dkr.ecr.us-east-1.amazonaws.com
    http-response replace-value Location ^(https://1234567890.dkr.ecr.us-east-1.amazonaws.com)(.*)$ https://example.com\2
[...]

PR #836, opened 2024-06-26
use the distribution protocol to determine account id and region

nicks:
*Issue #, if available:*
https://github.com/awslabs/amazon-ecr-credential-helper/issues/94

*Description of changes:*
use the distribution protocol to determine account id and region

this allows ecr-login to determine the correct
ECR instance, even when ECR is behind a custom domain.

### DIFF
--- a/ecr-login/ecr.go
+++ b/ecr-login/ecr.go
@@ -17,6 +17,9 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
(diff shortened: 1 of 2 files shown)


nicks:
ah, thinking about this, i'm not sure it's a good idea. works ok if you trust the registry. but if you use this with a malicious registry, this would allow it to steal creds. need to think on this a bit more. :thinking:

PR #873, opened 2024-09-18
Update Amazon Linux's distribution to v0.9.0

mysteriouspants:
*Issue #, if available:* N/A

*Description of changes:* `git merge v0.9.0` followed by updating the spec file to match the version of the source.


amazon-ecr-credential-helper.spec
```diff
@@ -139,6 +139,8 @@ install -D -m 0644 \
 rm -rf %{buildroot}
 
 %changelog
+* Wed 18 Sep 2023 Christopher R. Miller <milrchr@amazon.com> - 0.9.0-1
+- Update to v0.9.0
```
swagatbora90, line 143: may be add some details here?

swagatbora90 (commented):
Minor comment about changelog. Looks good to me otherwise.

amazon-ecr-credential-helper.spec
mysteriouspants: Yeah, we don't generally do that elsewhere in the distribution, but it's been consistently done for this package. I've just copy-pasted the items from https://github.com/awslabs/amazon-ecr-credential-helper/releases which appears to have been the convention previously followed.

swagatbora90 (approved):

Issue #249, opened 2020-12-17
Add support for mapping a registry to a specific AWS_PROFILE


micahhausler:
In the EKS Distro, we use ecr-credential-helper with [buildkit](https://github.com/moby/buildkit) to pull base images and push images, but we often have base images in one account, and need to push images to a separate account. Rather than using [ECR repository policies](https://docs.aws.amazon.com/AmazonECR/latest/userguide/repository-policies.html) to grant a single IAM role permission to do a in-account pull and a cross-account push, we use have a separate role our build job can assume to push cross-account to ECR public. I know ECR Public is already requested (#248), but simply adding support won't solve our use case.

ECR public doesn't (yet?) support repository policies, and we want our builder process to use a separate IAM role for ECR private pulls than the one used to push to ECR public.

Our current docker config looks like the following:
```json
{
    "credHelpers": {
        "111122223333.dkr.ecr.us-west-2.amazonaws.com": "ecr-login",
        "222233334444.dkr.ecr.us-west-2.amazonaws.com": "ecr-login"
    }
}
```

because using `{"credsStore": "ecr-login"}` swallows any token you try to save to the config file.

Right now, we have a [gross hack](https://github.com/aws/eks-distro/blob/cbc41df87455e54bf4580698674aa5ef976f0299/ecr-public/get-credentials.sh) that saves the ecr public credential in the docker config.
```bash
# Token expires after 12 hours
TOKEN=$(aws --region us-east-1 ecr-public get-authorization-token \
    --output=text \
    --query 'authorizationData.authorizationToken')

DOCKER_CONFIG=${DOCKER_CONFIG:-"~/.docker"}
mkdir -p $DOCKER_CONFIG
if [ ! -f $DOCKER_CONFIG/config.json ]; then
    echo '{}' > $DOCKER_CONFIG/config.json
fi
# Gross, abominable hack until ecr-credential-helper supports public ECR
# See https://github.com/awslabs/amazon-ecr-credential-helper/issues/248
cat $DOCKER_CONFIG/config.json | \
    jq --arg TOKEN $TOKEN '.auths."public.ecr.aws".auth = $TOKEN' > $DOCKER_CONFIG/config.json.new
cp $DOCKER_CONFIG/config.json.new $DOCKER_CONFIG/config.json
```

This allows buildkit to pull from one account and push from another in the same command.
```bash
BASE_IMAGE=111122223333.dkr.ecr.us-west-2.amazonaws.com/something/something
[...]

PR #894, opened 2024-11-03
Configuration file to map AWS Profiles with particular registries

callumcrossley12:
*Issue #, if available:* https://github.com/awslabs/amazon-ecr-credential-helper/issues/249

*Description of changes:*

The primary change is to add a new `registryConfig` file to map specific registries to use specific AWS_PROFILES. In general this is useful for environments where different profiles are used across accounts (i.e. for different AWS registries or permissions). I investigated using workarounds like credential wrappers but figured it may be more appropriate (particularly for us Windows users) to implement it in the tool. 

The general flow is:

1. Try to fetch the `registryConfig` file from disk
2. If it exists, try to fetch the profile for a specific registry with pattern
3. If profile is provided and is valid, use it for getting the auth token

If the file doesn't exist or the registry isn't mapped, it will fallback to using the default AWS configuration flow.

The file is simply a mapping of registries patterns to profiles. This could be extended in the future to include additional registry configuration if needed. Example file:

```yaml
registryConfigs:
  - registry: "123456789000.dkr.ecr.ap-southeast-2.amazonaws.com"
    config:
      profile: "some_profile"
  - registry: "987654321000.dkr.ecr.us-east-1.amazonaws.com"
    config:
      profile: "another_profile"
  - registry: "*.us-east-1.amazonaws.com"
    config:
      profile: "fallback_profile"
```

I've also added some minor changes to make Windows development a bit more friendly:
- Ensure `.sh` files use LF line endings
- Add `.vscode` to the `.gitignore`

This is both my first time raising a PR for an Opensource project and working with Go so feedback is appreciated.


ecr-login/ecr.go
```diff
@@ -115,13 +115,28 @@ func (self ECRHelper) Get(serverURL string) (string, string, error) {
 		return "", "", credentials.NewErrCredentialsNotFound()
 	}
 
+	profile, err := config.GetRegistryProfile(serverURL)
```
callumcrossley12, line 118: It may be overkill to try fetch the config file each time. I could potentially introduce a new env var (e.g. `AWS_ECR_USE_REGISTRY_CONFIG`) to determine whether to fetch?

ecr-login/ecr.go
```diff
@@ -115,13 +115,28 @@ func (self ECRHelper) Get(serverURL string) (string, string, error) {
...
+		self.logger.
+			WithError(err).
+			WithField("AWS profile", profile).
+			Errorf("Error fetching profile from config for repository %v", serverURL)
+		return "", "", credentials.NewErrCredentialsNotFound()
+	}
+
 	var client api.Client
 	if registry.FIPS {
 		client, err = self.clientFactory.NewClientWithFipsEndpoint(registry.Region)
```
callumcrossley12, line 129: Is there a scenario where we want to explicit use an explicit profile for Fips endpoints?

ecr-login/config/registry/registry.go
```diff
@@ -0,0 +1,93 @@
...
+)
+
+// Function to determine the RegistryConfigPath
+func getRegistryConfigPath() string {
+	// Get the path from the environment variable
+	path := os.Getenv(ENV_AWS_ECR_REGISTRY_CONFIG_PATH)
+	if path == "" {
+		expandedPath, err := os.UserHomeDir()
+		if err != nil {
+			expandedPath = "." // Fallback to the current directory if home directory cannot be resolved
```
dougbyrne, line 36: This could lead to some unexpected behavior. I would expect that the program proceed as if there is no config file in this case. Happy to be told that this is customary and expected though.

README.md
```diff
@@ -293,6 +298,31 @@ The credentials must have a policy applied that
...
+    profile: "<AWS_PROFILE_NAME>"
+```
+
+For example:
+
+```yaml
+registryConfigs:
+  123456789000.dkr.ecr.ap-southeast-2.amazonaws.com:
+    profile: "Profile1"
+  987654321000.dkr.ecr.ap-southeast-2.amazonaws.com:
```
dougbyrne, line 323: Do prefix wildcards work here? I usually have one profile per account ID, but it would be shared across all regions.

universam1:
Hi @coderbirju we are also missing the support for this functionality, especially in conjunction with Podman support. Would it be possible to land this feature please?

Issue #916, opened 2025-01-03
How to use this helper on windows?


marceliwac:
Hello,

I've spent about a day on debugging the issue now and I can't seem to be able to get the credentials helper to work on my Windows Server 2022 machine. Crucially, the setup uses the docker installation scripts from [here](https://learn.microsoft.com/en-us/virtualization/windowscontainers/quick-start/set-up-environment?tabs=dockerce#windows-server-1) and is **set up to run the windows containers**. I am **not** using Docker Desktop or WSL. 

I might be missing something obvious, so apologies if that's the case, but it seems that no matter what I try, I keep getting the following error:

```error
Error response from daemon: Head "https://123456789.dkr.ecr...": no basic auth credentials
```

## Steps I followed

1. Downloaded the pre-built binary from Releases page, as instructed.
2. Placed the binary (docker-credential-ecr-login.exe) in `C:\` (which is in PATH, running "docker-credential-ecr-login" in powershell shows the correct error message, see below).
3. Configured the `~\.aws\credentials` file to a standard format (and confirmed it works with aws cli).
4. Configured the `~\.docker\config.json` file to several different formats (see below) and verified that it is indeed the file used for docker configuration (basic auth credentials are placed there when executing aws cli commands, malformed json results in errors when trying to pull the image etc.).
5. Ran `docker pull 12345678.dkr.ecr..../myimage` to no avail, each time the "no basic auth credentials comes up". 

#### credsStore
```json
{
  "credsStore": "ecr-login"
}
```

#### credsStore (with empty auths)
```json
{
  "auths": {
    "12345678.dkr.ecr.eu-west-1.amazonaws.com": {}
  },
  "credsStore": "ecr-login"
}
```

#### credHelpers
```json
{
  "credHelpers": {
    "12345678.dkr.ecr.eu-west-1.amazonaws.com": "ecr-login"
  }
}
```

#### credHelpers (with empty auths)
```json
{
  "auths": {
    "12345678.dkr.ecr.eu-west-1.amazonaws.com": {}
  },
  "credHelpers": {
    "12345678.dkr.ecr.eu-west-1.amazonaws.com": "ecr-login"
  }
}
```

#### credsStore and credHelpers
```json
{
  "credHelpers": {
    "12345678.dkr.ecr.eu-west-1.amazonaws.com": "ecr-login"
  },
  "credsStore": "ecr-login"
}
```

#### credsStore and credHelpers (with empty auths)
```json
{
[...]

Issue #81, opened 2017-10-18
Windows 2016


goffinf:
I have been successfully using this helper on Linux (RHEL) for quite a while.

We have just started to configure AWS EC2 instances running Windows 2016 (configured for docker) and these also need to be able to pull and push images to/from ECR.

The Amazon docs for this helper here: https://aws.amazon.com/blogs/compute/authenticating-amazon-ecr-repositories-for-docker-cli-with-credential-helper/ state that it is certainly usable for Windows. We have followed the instructions for creating a Windows compatible binary.

The question that I have though is *where* should I locate the .docker/config.json file (we are using an IAM role for the AWS creds).

Most blogs, articles and such talk about ~/.docker/config.json but where would that be on my Windows 2016 instance when the CLI is invoked NOT by a user logged into the instance but via automation, such as a Jenkins job or a docker pull requested from a docker container management tool when deploying a stack ?

For Linux, we might put the .docker/config.json under the ec2-user home directory or even under /root, but where is the corresponding location for Windows ?

Regards

Fraser.

samuelkarp:
> [W]here would that be on my Windows 2016 instance when the CLI is invoked NOT by a user logged into the instance but via automation, such as a Jenkins job

You'd put it in the `~/.docker/config.json` of the user under which automation is running.

digitalfiz:
@goffinf Looks like its in `C:/Users/Username/.docker/machine/default/config.json` from a quick google search leading to the SO: https://stackoverflow.com/questions/37112422/docker-config-file-location-on-windows-to-e-g-enable-insecure-registry-dock

Tim-Co:
For the system account you need to put it here: `C:\Windows\system32\config\systemprofile\.docker\config.json`

samuelkarp:
Closing, as this question appears to be answered.

PR #929, opened 2025-01-21
docs: Add details to configuration section for Windows.

marceliwac:
Related to issues: #916 #81

Expand on where the configuration settings should reside. Distinguish between the user and root/system contexts and the corresponding directories for configurations files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,8 +209,21 @@ Docker to work with the helper.
 
 There is no need to use `docker login` or `docker logout`.
 
-Place the `docker-credential-ecr-login` binary on your `PATH` and set the
-contents of your `~/.docker/config.json` file to be:
+Place the `docker-credential-ecr-login` binary on your `PATH`.
+On Windows, depending on whether the executable is ran in the User or System context, the corresponding `Path` user or system variable needs to be used.
+
+Following that the configuration for the docker client needs to be updated in `~/.docker/config.json` to use the **ecr-login** helper.
+Depending on the operating system and context under which docker client will be executed, this configuration can be found in different places.
+  
+On Linux systems:
+- `/home/<username>/.docker/config.json` for **user** context
+- `/root/.docker/config.json` for **root** context
+  
+On Windows:
+- `C:\Users\<username>\.docker\config.json` for **user** context
+- `C:\Windows\System32\config\systemprofile\.docker\config.json` for the **SYSTEM** context
+
+Set the contents of the file to the following:
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -209,8 +209,21 @@ Docker to work with the helper.
 
 There is no need to use `docker login` or `docker logout`.
 
-Place the `docker-credential-ecr-login` binary on your `PATH` and set the
-contents of your `~/.docker/config.json` file to be:
+Place the `docker-credential-ecr-login` binary on your `PATH`.
```
sondavidb, line 212: Docker actually has a builtin bin folder that should(?) be added to your PATH automatically in `C:\Program Files\Docker\Docker\resources\bin` (though I'm not certain this is the case). We should probably suggest putting it in there instead.

marceliwac: Thanks for reviewing the changes @sondavidb.

On Windows Server, when installed via Microsoft's [install script](https://learn.microsoft.com/en-us/virtualization/windowscontainers/manage-docker/configure-docker-daemon#install-docker), docker binaries are actually placed under `C:\Windows\System32`. This directory is on the `Path` env variable for the System, but not the default user (Administrator), but these should be merged when accessed under user context. In either case, I think that placing the docker-credential-ecr-login binary next to docker's binaries would make sense in the long run.

I think this could be addressed in a separate PR, perhaps one that provides some installation scripts for platforms that don't include built-in package manager support (such as Windows). Alternatively, I would love to see the support for this package in Chocolatey, which could also handle the configuration during installation.

To reflect on where this PR comes from, the particular pain of mine was ensuring that the config files had the right contents and were saved in the correct place. There is a lot of misleading information both within the issues of this repository and other websites (e.g. StackOverflow), where the suggestions to overcome the issue of the helper not being used are providing config with `"credStore"` rather than `"credsStore"` (which changes things but not for the better - docker just disregards these values since it's not looking for them under that key). It took some time to figure out that the issue was related to the system context on Windows using a different directory for the docker config. Perhaps it's my ignorance and lack of understanding of "the Windows way". At any rate, while I can't imagine that building Windows Docker Containers on Windows Server that isn't using Docker Desktop is a popular use-case, this small PR that clarifies the config documentation should hopefully help alleviate similar issues on other platforms.

sondavidb (approved):

README.md
sondavidb: > I think this could be addressed in a separate PR, perhaps one that provides some installation scripts for platforms that don't include built-in package manager support (such as Windows). Alternatively, I would love to see the support for this package in Chocolatey, which could also handle the configuration during installation.

PRs are always welcome, particularly since it seems you're doing work in Windows, which we have less experience in. (Incidentally I use Windows but almost all of my dev work is on remote Linux machines because native Windows development was a pain 🙂)

sondavidb:
Thanks for this!

PR #970, opened 2025-05-13
chore: remove dollar signs from scripts in readme

j-cordial:
GitHub has a handy feature where readers can click a copy button to copy the context of a code block. In this case, when you copy these blocks - the dollar sign is included. 

We can do better. Having the language declared at the top of the code block should suffice.

*Issue #, if available:*

*Description of changes:*
- removes dollar signs from bash/shell code blocks in the root Readme.


austinvazquez (approved):
LGTM, thanks for the contribution.

PR #1092, opened 2026-06-02
Propagate context.Context through ECR credential helper

matejvasek:
## Summary

- Add `WithContext(ctx)` option to `NewECRHelper()` that propagates a caller-supplied `context.Context` through all internal layers (factory, client, cache) to the actual AWS SDK calls
- Replace all 6 `context.TODO()` call sites with the propagated context, making ECR operations cancellable and deadline-aware
- Default to `context.Background()` when `WithContext` is not used, preserving existing behavior

## Motivation

This change was prompted by difficulties encountered in [knative/func#3864](https://github.com/knative/func/pull/3864), where the ECR credential helper's use of `context.TODO()` made it impossible to cancel or time out hanging network requests. When AWS SDK calls (config loading, `GetAuthorizationToken`, credential retrieval) stall, the caller has no way to interrupt them because the context is hardcoded.

The `credentials.Helper` interface from `docker-credential-helpers` has fixed method signatures with no context parameter, so the context is stored in the `ECRHelper` struct and passed internally.

## Breaking changes

The `Client`, `ClientFactory` interfaces and `BuildCredentialsCache` function now require a `context.Context` first parameter. External consumers of these APIs will need to update call sites.

However I believe most people use the API at the `credentials.Helper` level.

## Note on panic in factory methods

`NewClientWithDefaults` and `NewClientFromRegion` call `config.LoadDefaultConfig(ctx, ...)` and `panic` on error, while `NewClientWithFipsEndpoint` properly returns the error to the caller. This was a latent issue before this change — with `context.TODO()`, `LoadDefaultConfig` could only fail due to malformed config files or missing credential providers, which are arguably unrecoverable. But now that we pass a real context, `LoadDefaultConfig` can also fail for a routine, expected reason: context cancellation. It makes network calls (IMDS for region/credential discovery, STS for token exchange, etc.) that all respect the context. A caller that sets a short deadline via `WithContext` will get a `panic` instead of a graceful error if e.g. IMDS is slow. Note that this is a non-issue for default usage — without `WithContext`, the helper uses `context.Background()` which is never cancelled, preserving the existing behavior. The panic can only be triggered when a caller explicitly opts in to a cancellable context.

Two possible follow-up fixes:

1. **Return `(Client, error)` from all factory methods** — align `NewClientWithDefaults` and `NewClientFromRegion` with `NewClientWithFipsEndpoint`, which already returns an error. This is the smaller change since the `ClientFactory` interface is already being modified in this PR, but it pushes error handling to every call site.

2. **Lazy config loading** — defer `LoadDefaultConfig` until the first API call (`GetCredentials`, `ListCredentials`), where the error can be returned through the existing `error` return value. Factory methods would just capture parameters and the `defaultClient` would load config on first use (via `sync.Once` or similar). This keeps construction infallible and avoids further interface changes, at the cost of added internal complexity.

### DIFF
--- a/ecr-login/api/client_test.go
+++ b/ecr-login/api/client_test.go
@@ -18,11 +18,15 @@ import (
 	"encoding/base64"
 	"errors"
 	"fmt"
+	"net/http"
+	"net/http/httptest"
 	"net/url"
+	"os"
 	"testing"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
+	awscreds "github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/service/ecr"
 	ecrtypes "github.com/aws/aws-sdk-go-v2/service/ecr/types"
 	"github.com/aws/aws-sdk-go-v2/service/ecrpublic"
@@ -291,7 +295,7 @@ func TestGetAuthConfigSuccess(t *testing.T) {
 		compareAuthEntry(t, actual, authEntry)
 	}
 
-	auth, err := client.GetCredentials(proxyEndpoint)
+	auth, err := client.GetCredentials(context.Background(), proxyEndpoint)
 	assert.Nil(t, err)
 	assert.Equal(t, auth.Username, expectedUsername)
 	assert.Equal(t, auth.Password, expectedPassword)
@@ -320,7 +324,7 @@ func TestGetAuthConfigNoMatchAuthorizationToken(t *testing.T) {
 
 	credentialCache.GetFn = func(_ string) *cache.AuthEntry { return nil }
 
-	auth, err := client.GetCredentials(proxyEndpoint)
+	auth, err := client.GetCredentials(context.Background(), proxyEndpoint)
 	assert.NotNil(t, err)
 	assert.Nil(t, auth)
 }
@@ -349,7 +353,7 @@ func TestGetAuthConfigGetCacheSuccess(t *testing.T) {
 		return authEntry
 	}
 
-	auth, err := client.GetCredentials(proxyEndpoint)
+	auth, err := client.GetCredentials(context.Background(), proxyEndpoint)
 	assert.Nil(t, err)
 	assert.Equal(t, auth.Username, expectedUsername)
 	assert.Equal(t, auth.Password, expectedPassword)
@@ -403,7 +407,7 @@ func TestGetAuthConfigSuccessInvalidCacheHit(t *testing.T) {
 		compareAuthEntry(t, actual, authEntry)
 	}
 
-	auth, err := client.GetCredentials(proxyEndpoint)
+	auth, err := client.GetCredentials(context.Background(), proxyEndpoint)
 	assert.Nil(t, err)
 	assert.Equal(t, auth.Username, expectedUsername)
 	assert.Equal(t, auth.Password, expectedPassword)
@@ -432,7 +436,7 @@ func TestGetAuthConfigBadBase64(t *testing.T) {
 
 	credentialCache.GetFn = func(_ string) *cache.AuthEntry { return nil }
 
-	auth, err := client.GetCredentials(proxyEndpoint)
+	auth, err := client.GetCredentials(context.Background(), proxyEndpoint)
 	assert.NotNil(t, err)
 	t.Log(err)
 	assert.Nil(t, auth)
@@ -455,7 +459,7 @@ func TestGetAuthConfigMissingResponse(t *testing.T) {
 
 	credentialCache.GetFn = func(_ string) *cache.AuthEntry { return nil }
 
-	auth, err := client.GetCredentials(proxyEndpoint)
+	auth, err := client.GetCredentials(context.Background(), proxyEndpoint)
 	assert.NotNil(t, err)
 	t.Log(err)
 	assert.Nil(t, auth)
@@ -478,7 +482,7 @@ func TestGetAuthConfigECRError(t *testing.T) {
 
 	credentialCache.GetFn = func(_ string) *cache.AuthEntry { return nil }
 
-	auth, err := client.GetCredentials(proxyEndpoint)
+	auth, err := client.GetCredentials(context.Background(), proxyEndpoint)
 	assert.NotNil(t, err)
 	t.Log(err)
 	assert.Nil(t, auth)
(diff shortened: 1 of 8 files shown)


matejvasek:
I fixed the potential panic by using `Return (Client, error)` in separate commit.

matejvasek:
OK I did some small tweaks, now it should be good to go.

matejvasek:
Please review.

ecr-login/api/client_test.go
```diff
@@ -954,3 +958,87 @@ func TestECRPublicClientWrapper(t *testing.T) {
...
+	// Direct the SDK's credential chain to our hanging server and
+	// disable all other credential sources.
+	t.Setenv("AWS_CONTAINER_CREDENTIALS_FULL_URI", server.URL+"/creds")
+	t.Setenv("AWS_EC2_METADATA_DISABLED", "true")
+	t.Setenv("AWS_CONFIG_FILE", os.DevNull)
+	t.Setenv("AWS_SHARED_CREDENTIALS_FILE", os.DevNull)
+	t.Setenv("AWS_ACCESS_KEY_ID", "")
+	t.Setenv("AWS_SECRET_ACCESS_KEY", "")
+
+	ctx, cancel := context.WithTimeout(context.Background(), 200*time.Millisecond)
```
sondavidb: 200ms is small but not sure how sufficiently small. I'd feel a little more comfortable with 1ms or so.

Alternatively, we can make it completely deterministic by calling `cancel()` right before the `client.GetCredentials` call and then check for `context.Canceled` instead.

ecr-login/api/factory.go
```diff
@@ -34,11 +35,11 @@ type Options struct {
 
 // ClientFactory is a factory for creating clients to interact with ECR
 type ClientFactory interface {
```
sondavidb, line 37: While we have the hood open, wonder if we should just make all the return types to be `(Client, error)`. Seems confusing to have a seemingly random correlation between `NewClientxxx` and the return type having or not having an error.

ecr-login/api/client_test.go
```diff
@@ -954,3 +958,87 @@ func TestECRPublicClientWrapper(t *testing.T) {
 		assert.NoError(t, err)
 	})
 }
+
+func TestNewClientFromRegionContextCancel(t *testing.T) {
```
sondavidb: Can we combine this with the below test and test all of the NewClient calls?

ecr-login/ecr.go
```diff
@@ -29,6 +30,9 @@ import (
 var notImplemented = errors.New("not implemented")
 
 type ECRHelper struct {
+	// ctx is stored because the credentials.Helper interface methods
+	// do not accept a context parameter.
```
sondavidb, line 34: I don't really like having to store the context within the `ECRHelper` struct (Go generally suggests against it, [ref](https://go.dev/blog/context-and-structs)) but it does seem to be the best way to get around Docker's API not having one since we don't have control over the API.

It does imply that this struct has a lifecycle of sorts, which isn't what we're trying to do. I think the comment clarifies the intent well enough but feel free to amend as needed to drive home this point. (Happy with no change to the comment as well)

sondavidb (commented):
Hey there, thanks for the contribution, we do appreciate it 🙂 

Approach seems fine, have a couple of concerns but shouldn't be huge changes.

ecr-login/ecr.go
matejvasek: I do not like it either but I don't think there are much better ways. In the end the context would end up in **some** struct **somewhere** (or god forbid global variable).


We could kinda make it **someone else problem** by accepting custom `http.RoundTripper` instead of `context.Context`. Caller then could pass their implementation of `http.RoundTripper` that would internally contained the context and cancelled http requests accordingly. But again the context is member of some struct. One problem is that  it only could cancel HTTP requests not other long running operations. However I believe the most hang ups happens because of HTTP requests. Also to be able to do this the AWS SDK would have to support custom `RoundTrippers`.


Some libraries as Docker Client lib or go-containerresitry lib accept custom transport (`http.RoundTripper`). It  is really useful and neat option. You can use it to affect TLS validation, or use custom **dialer** function -- dial unix socket instead of tcp (or even do some tunneling via ssh). 


Only real clean solution would be to create a context aware helper interface. All the methods of the Helper interface would have to accept a context as their (fist) parameter.

matejvasek: Claude say

>Yes, the AWS SDK v2 fully supports it [custom transport]. The aws.Config has an HTTPClient field that accepts any type implementing
> type HTTPClient interface { Do(*http.Request) (*http.Response, error); }

if we could trust it.

Again this [custom transport] would be a lot more work with uncertain result.

matejvasek:
OK, I pushed requested changes.

ecr-login/api/client_test.go
matejvasek: I changed as requested to use explicit `cancel()`.

>200ms is small but not sure how sufficiently small.

Well the testing server hangs indefinitely so even 10 second would have been all right I think.

ecr-login/api/factory.go
matejvasek: Done.

ecr-login/api/client_test.go
matejvasek: Done.

ecr-login/api/client_test.go
matejvasek: >200ms is small but not sure how sufficiently small.

or you were worried about test execution times?

ecr-login/ecr.go
sondavidb: I like the idea of the custom client but I agree that it's too much work given the scope of what this is trying to achieve. I think we can put a pin on this and rethink it down the line, and if we make a change we should do it outside of this PR.

sondavidb (approved):
LGTM, thanks! Let me see if I can get 1 more reviewer

ecr-login/api/client_test.go
sondavidb: No you're right, forgot the test server is an indefinite hang so it doesn't matter. I think the cancel call is still a little cleaner, looks like you've made this changes so no more concerns from me.

ayush-panta (approved):

sondavidb:
Oh last thing, do you mind squashing your commits @matejvasek ?

matejvasek:
Squashed.

